PR #40, opened 2026-02-22
feat: add Kilo CLI platform support

Xintong120:
- Add kilo to AITool, TemplateDir, and CliFlag types
- Add AI_TOOLS configuration for Kilo CLI (.kilocode directory)
- Add getKiloTemplatePath() and readKiloFile() in extract.ts
- Create src/templates/kilo/ with Trellis commands
- Create src/configurators/kilo.ts for platform configuration
- Add collectTemplates() for update tracking
- Add --kilo CLI option
- Add kilo to InitOptions type
- Add tests for kilo template functions

Note: Kilo does not support hooks like Claude Code, so auto-injection is not available. Users need to manually trigger /trellis:start command. This is the same experience as Cursor and Codex platforms.

### DIFF
--- a/src/templates/kilo/commands/trellis/brainstorm.md
+++ b/src/templates/kilo/commands/trellis/brainstorm.md
@@ -1,0 +1,476 @@
+# Brainstorm - Requirements Discovery (AI Coding Enhanced)
+
+Guide AI through collaborative requirements discovery **before implementation**, optimized for AI coding workflows:
+
+* **Task-first** (capture ideas immediately)
+* **Action-before-asking** (reduce low-value questions)
+* **Research-first** for technical choices (avoid asking users to invent options)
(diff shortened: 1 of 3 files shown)


taosu0216:
Thanks for the contribution! A few items need attention before merging:
Missing from the checklist (see .trellis/spec/backend/platform-integration.md):
1. Step 6 - Python scripts: cli_adapter.py needs kilo added to Platform type, config_dir_name, and detect_platform(). The pattern for 4 existing platforms is already there — just follow it.

Other issues:
migrate-specs.md is an empty file (0 bytes). 
Plan: We'll merge feat/codex into main first (which adds brainstorm command + codex platform). After that, please rebase onto the latest main and address the above. That way you'll have the full codebase to reference.

PR #147, opened 2026-04-04
fix(opencode): make dispatch wait for child tasks

1508324011:
## Summary
- switch the OpenCode dispatch instructions to synchronous child `Task(...)` calls instead of background `TaskOutput` polling
- add regression coverage so the OpenCode template cannot quietly reintroduce `run_in_background: true` child phases
- document the OpenCode platform rule that background task status is not the final completion signal for child sessions

## Testing
- pnpm --filter @mindfoldhq/trellis test -- regression.test.ts
- pnpm --filter @mindfoldhq/trellis build

### DIFF
--- a/.opencode/agents/dispatch.md
+++ b/.opencode/agents/dispatch.md
@@ -71,6 +71,10 @@ Execute each step in `phase` order.
 
 > Hook will auto-inject all specs, requirements, and technical design to subagent context.
 > Dispatch only needs to issue simple call commands.
+>
+> **OpenCode dispatch rule**: Call subagents synchronously (`run_in_background: false`).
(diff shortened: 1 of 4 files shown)


1508324011:
Follow-up note: after more testing, I no longer think this PR should be described as a complete fix.

It still looks useful as a mitigation because it removes one error-prone dispatch pattern in the OpenCode template.

But in longer conversations, the underlying background completion semantics can still be wrong even when the prompt says to use synchronous child calls. So this PR is better framed as a patch / risk reduction step, while the root cause is likely in the lower-level `TaskOutput` / `background_output` completion mechanism (or a wrapper below Trellis).

I opened a follow-up issue to track that root problem separately.

taosu0216:
good job

PR #160, opened 2026-04-12
fix(session-start): remove redundant start.md injection

### LINKED ISSUE #154: SessionStart hook output exceeds Claude Code's additionalContext limit, silently truncating task state
## Observation

On a fresh `trellis init --claude` install (verified against v0.4.0-beta.8), the `session-start.py` hook produces `additionalContext` that is consistently over 30 KB. Claude Code appears to truncate `additionalContext` at around 20 KB, emitting only a ~2 KB preview plus a fallback file path. In practice the AI does not read the fallback file unless explicitly prompted, so anything past the 2 KB preview is effectively invisible — **including `## ACTIVE TASKS`, `## CURRENT TASK`, and `## MY TASKS`**.

End result: the session-start context injection silently breaks for any project whose combined size pushes it past ~20 KB, which is most non-trivial projects. The developer sees "task state not being surfaced" without any error.

## Data

Measurements from the hook output on two projects (one clean vanilla install, one lightly used):

| Segment | Project A (vanilla) | Project B (light use) |
|---|---:|---:|
| `<session-context>` | ~150 B | ~150 B |
| `<current-state>` (tasks, git, journal) | 1.6 KB | 1.7 KB |
| `<workflow>` **(full `workflow.md`)** | **12.5 KB** | **16.6 KB** |
| `<guidelines>` (spec index files) | 6.3 KB | 5.0 KB |
| `<instructions>` **(full `start.md`)** | **11.0 KB** | **11.0 KB** |
| **total `additionalContext`** | **~31 KB** | **~36 KB** |

Both are over the observed ~20 KB truncation threshold. The two large blocks marked ★ in the table above account for **75%+ of the payload**.

## How I verified the truncation behavior

Using `claude -p --session-id <uuid>` to spawn a fresh Claude Code instance inside the project and asking it to report what it actually received:

- Fresh instance replies: "additionalContext was truncated; only a ~2 KB preview was available; saved to `tool-results/hook-<id>-additionalContext.txt`"
- The Active Tasks / Current Task sections are **not visible** in the visible preview
- The instance does not proactively read the fallback file

After reducing the hook output to well under 20 KB (see "Possible approaches" below), a fresh Claude instance reports:
- "No truncation. I can see 3 active tasks, the current task name, and the full workflow section index."

## Possible approaches (not prescriptive)

I experimented with a couple of out-of-tree patches on my own projects and both worked well; sharing here as ideas for upstream discussion, not as a fait-accompli solution.

**Approach A — Dynamic TOC for `workflow.md`**

Instead of reading `workflow.md` verbatim into `<workflow>`, `session-start.py` scans the file at session-start time and emits a compact table of contents with line ranges for each `##` section. Claude navigates by line range and reads specific sections on-demand via `Read offset+limit`. A compact form:

```
<workflow>
`.trellis/workflow.md` — read on-demand by section (use Read with offset+limit):
- L7-18 (12L): Table of Contents
- L19-88 (70L): Quick Start
- L89-150 (62L): Workflow Overview
- L151-202 (52L): Session Start Process
- L203-240 (38L): Development Process
- L241-270 (30L): Session End
- L271-337 (67L): File Descriptions
- L338-368 (31L): Best Practices
- L369-406 (38L): Quick Reference
- L407-420 (14L): Summary
</workflow>
```

Typical size: ~500–700 bytes instead of 12–16 KB. Claude still has full content access — it just reads it lazily.

**Approach B — Remove `start.md` from session-start entirely**

`start.md` is the `/trellis:start` slash command definition. Claude Code already expands it as a prompt when the user types the command. Injecting it at session-start as well is duplicate work: the user either types the command (in which case they get the full content on demand) or they don't (in which case the 11 KB is unused ballast). Other slash commands (`brainstorm`, `finish-work`, `check`, `record-session`) are not pre-injected — so this would restore symmetry.

Replace the current `<instructions>` block with a 1-line pointer like:

```
<instructions>
For task startup workflow, type the slash command `/trellis:start`. It expands the full guide on demand. Other slash commands (brainstorm / finish-work / check / record-session) follow the same pattern — not pre-injected here.
</instructions>
```

Typical savings: ~11 KB.

**Combined result on my measurement projects**

With both Approach A and B applied:

| Project | Before | After |
|---|---:|---:|
| A (vanilla) | 31 KB | **9.3 KB** |
| B (light use) | 36 KB | **8.4 KB** |

Both drop well under the truncation threshold. Fresh `claude -p` instances in both projects confirm they now see the full `<current-state>` and `<workflow>` TOC, with room to spare (~11 KB headroom before hitting the limit).

## Reproduction

```bash
mkdir /tmp/trellis-repro && cd /tmp/trellis-repro && git init
npx @mindfoldhq/trellis@beta init --claude -u test-user -y

# Measure session-start output size:
echo '{"session_id":"t","source":"startup"}' \
  | python3 .claude/hooks/session-start.py \
  | python3 -c 'import sys,json; d=json.load(sys.stdin); ctx=d["hookSpecificOutput"]["additionalContext"]; print(f"{len(ctx.encode(\"utf-8\"))} bytes")'
```

Typical vanilla output: **30+ KB** on a repo with a handful of tasks and spec files.

## Why I'm raising this as an issue, not a PR

This is a design-space question more than a clear bug:

1. Is the truncation behavior I'm seeing a Claude Code platform limit, a trellis design assumption, or both?
2. If workflow.md / start.md pre-injection is intentional, what's the rationale? (I'd like to understand before proposing a change — I might be missing a reason.)
3. If reducing the payload is welcome, which approach fits better — dynamic TOC, symmetry with other slash commands, a config toggle, all three?

I'm happy to turn either or both approaches into a PR if there's interest, but I'd rather discuss the direction first so the work doesn't go to waste.

## Environment

- trellis: 0.4.0-beta.8 (also observed on beta.9)
- Claude Code: 4.6
- OS: macOS 14 (Darwin 23.6)
- Python: 3.11
- node: 20.x

## Related

- Search turned up no existing issues for `SubagentStop` / `truncat` / size-related SessionStart discussion
- Related fixes I'm opening separately: #152 (ralph-loop field names), #153 (migration task assignee parsing)

21nak:
The /trellis:start slash command is already expanded on demand by the
platform (Claude Code, Codex, iFlow, OpenCode), so pre-injecting the
11 KB start.md into every session wastes ~11 KB of the additionalContext
budget for no benefit and creates an inconsistency with other commands
that do not receive the same treatment.

Remove the <instructions> block from all session-start hooks and update
the <ready> message to no longer reference it.

Affected hooks: claude, codex, iflow (Python) and opencode (JS)
across both live hooks and src/templates.

Fixes mindfold-ai/Trellis#154 (Approach B)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/.claude/hooks/session-start.py
+++ b/.claude/hooks/session-start.py
@@ -365,20 +365,13 @@ Read and follow all instructions below carefully.
 
     output.write("</guidelines>\n\n")
 
-    output.write("<instructions>\n")
-    start_md = read_file(
-        claude_dir / "commands" / "trellis" / "start.md", "No start.md found"
-    )
-    output.write(start_md)
-    output.write("\n</instructions>\n\n")
-
     # Check task status and inject structured tag
     task_status = _get_task_status(trellis_dir)
     output.write(f"<task-status>\n{task_status}\n</task-status>\n\n")
 
     output.write("""<ready>
 Context loaded. Steps 1-3 (workflow, context, guidelines) are already injected above — do NOT re-read them.
-Start from Step 4. Wait for user's first message, then follow <instructions> to handle their request.
+Start from Step 4. Wait for user's first message, then follow the workflow to handle their request.
 If there is an active task, ask whether to continue it.
 </ready>""")
 
--- a/.codex/hooks/session-start.py
+++ b/.codex/hooks/session-start.py
@@ -193,21 +193,12 @@ Read and follow all instructions below carefully.
 
     output.write("</guidelines>\n\n")
 
-    # Inject start skill as instructions (Codex uses skills, not slash commands)
-    start_skill = codex_dir / "skills" / "start" / "SKILL.md"
-    if not start_skill.is_file():
-        start_skill = project_dir / ".agents" / "skills" / "start" / "SKILL.md"
-    if start_skill.is_file():
-        output.write("<instructions>\n")
-        output.write(read_file(start_skill))
-        output.write("\n</instructions>\n\n")
-
     task_status = _get_task_status(trellis_dir)
     output.write(f"<task-status>\n{task_status}\n</task-status>\n\n")
 
     output.write("""<ready>
 Context loaded. Steps 1-3 (workflow, context, guidelines) are already injected above — do NOT re-read them.
-Start from Step 4. Wait for user's first message, then follow <instructions> to handle their request.
+Start from Step 4. Wait for user's first message, then follow the workflow to handle their request.
 If there is an active task, ask whether to continue it.
 </ready>""")
 
--- a/.opencode/plugins/session-start.js
+++ b/.opencode/plugins/session-start.js
@@ -82,20 +82,9 @@ Read and follow all instructions below carefully.
 
   parts.push("</guidelines>")
 
-  // 5. Session Instructions - try both .claude and .opencode
-  let startMd = ctx.readFile(join(claudeDir, "commands", "trellis", "start.md"))
-  if (!startMd) {
-    startMd = ctx.readFile(join(opencodeDir, "commands", "trellis", "start.md"))
-  }
-  if (startMd) {
-    parts.push("<instructions>")
-    parts.push(startMd)
-    parts.push("</instructions>")
-  }
-
   // 6. Final directive
   parts.push(`<ready>
-Context loaded. Wait for user's first message, then follow <instructions> to handle their request.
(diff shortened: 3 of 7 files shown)


taosu0216:
Rebased onto main to clear the #159 conflict (pure text drift, no semantic change). Authorship preserved. Thanks

PR #163, opened 2026-04-14
fix(hooks): add Windows GBK encoding fix to statusline.py

xiangagou163:
修复claude code windows状态栏乱码显示windows乱码显示
修复后
image


taosu0216:
Thanks — nice catch. I added one follow-up commit to sync the same fix into `packages/cli/src/templates/claude/hooks/statusline.py` so new `trellis init` installs also pick it up (the live `.claude/hooks` and the template are separate copies).

PR #277, opened 2026-05-13
修复 Codex hook 的 UTF-8 读取模式

Tim-Devil:


Tim-Devil:
## Summary

This change makes the Codex workflow-state hook run Python in UTF-8 mode:

```json
"{{PYTHON_CMD}} -X utf8 .codex/hooks/inject-workflow-state.py"
```

## Motivation

On Windows, Python may decode stdin using the active ANSI code page instead of UTF-8. Codex passes hook input as UTF-8 JSON, so when a project path contains non-ASCII characters, such as Chinese characters, the `cwd` field can be decoded incorrectly. In that case, the Trellis root lookup fails and the hook exits successfully without injecting the expected `<workflow-state>` context.

Enabling Python UTF-8 mode makes the hook decode Codex input consistently and fixes the silent context-injection failure for these paths.

## Scope

This patch only updates the Codex hook template because this is the platform where we reproduced the issue. **Similar encoding problems may exist in other hook-capable terminals or platforms on Windows/non-UTF-8 environments, but this PR keeps the fix intentionally narrow until those cases are reproduced!!!**

taosu0216 (approved):
Confirmed: `-X utf8` correctly forces Python UTF-8 mode (PEP 540), fixes Windows cp936/cp1252 stdin decode errors on Chinese content. Python ≥ 3.7 supports the flag; trellis already requires ≥ 3.9. Merging this for codex Windows users now; will follow up to extend the same protection to claude/cursor/gemini/copilot/droid hooks.

PR #337, opened 2026-06-14
Migrate Trellis workflow to YAML manifest

yongchangwang-gt:
## Summary
- Replace the monolithic `.trellis/workflow.md` guide with `.trellis/workflow.yaml` plus `.trellis/workflow/**` markdown body files.
- Add shared workflow model loaders for Python/OpenCode/Pi runtime consumers, and update CLI init/update templates, hooks, specs, bundled skills, and regression tests.
- Add bilingual i18n support for the affected CLI/template surfaces.
- Bump `@mindfoldhq/trellis` to `0.5.21` and restore published `0.6.0-beta.22`, `0.6.0-beta.23`, and `0.6.0-rc.0` migration manifests so continuity checks stay green.

## Verification
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test`
- `node scripts/check-manifest-continuity.js` from `packages/cli`
- `node scripts/check-i18n.js --release` from `packages/cli`

## Previous targeted verification
- `python3 -m py_compile` on updated Python runtime/hooks/templates
- `node --check` on updated OpenCode files
- targeted CLI template/update/regression Vitest suites
- `pnpm --filter @mindfoldhq/trellis build`
- smoke-tested `get_context.py` phase rendering plus Codex workflow-state/session-start hooks


taosu0216:
Checking in on this after far too long — it has been open since June with no response from us, which is on the maintainers, not on you.

Where it stands today: `mergeable: CONFLICTING` against a `main` that has moved a long way since (we are on 0.6.15 now, with several platforms added and the workflow template reworked more than once). A 211-file migration to a YAML manifest touches almost exactly the surfaces that changed most.

Two honest options:

1. **You still want it.** Say so and tell us what you would need from us to make it viable — in particular whether the YAML manifest direction is something we should commit to before you spend time rebasing, because rebasing 211 files against four months of drift is not work worth doing on spec. We would owe you a decision on the direction first, and we should have given you that in June.

2. **It has run its course.** Also completely fine. We would close it as stale with the branch intact — nothing is lost, and it can be reopened any time.

If we do not hear back in a couple of weeks we will close it under option 2, purely as housekeeping rather than a verdict on the idea. Either way, sorry for leaving it unanswered this long.

PR #352, opened 2026-06-18
fix(channel): resolve .exe path from npm .cmd shim on Windows

### LINKED ISSUE #348: channel spawn fails on Windows — spawn("claude") ENOENT after Node.js CVE-2024-27980
# Issue: `channel spawn` fails on Windows — `spawn claude ENOENT`

## Environment
- Windows 10, Node v24.14.1 (nvm-windows), Trellis 0.6.2
- Claude Code 2.1.181 installed via npm globally

## Steps
```
trellis channel create test --type forum
trellis channel spawn test --agent check --as worker1
```

## Result
```
worker spawn failed: spawn claude ENOENT
```

## Cause

`supervisor.js` calls `spawn("claude", args, { shell: false })`.

On Windows, npm global CLI tools follow the standard `.cmd`-shim pattern: the
real `claude.exe` lives in `node_modules/` and only a `claude.cmd` wrapper sits
in PATH.  Since Node.js v18.20.2 (CVE-2024-27980), `spawn()` no longer resolves
`.cmd` files without `shell: true`, returning ENOENT.

This is not a `shell: true` fix — the args include `--append-system-prompt` with
agent/context content that shouldn't pass through `cmd.exe`.

## Proposed fix (see PR)

Resolve the real `.exe` path from the npm `.cmd` shim before spawning:

```ts
function resolveProviderPath(provider: string): string {
    if (process.platform !== "win32") return provider;
    try {
        const dirs = (process.env.PATH || "").split(path.delimiter);
        for (const dir of dirs) {
            const cmdFile = path.join(dir, provider + ".cmd");
            if (!fs.existsSync(cmdFile)) continue;
            const m = fs.readFileSync(cmdFile, "utf8").match(/"%dp0%\\(.+\.exe)"/i);
            if (m) {
                const exePath = path.join(dir, m[1]);
                if (fs.existsSync(exePath)) return exePath;
            }
        }
    } catch { }
    return provider;
}
```

Then in the spawn call: `spawn(resolveProviderPath(adapter.provider), args, { … })`.

- 1 file changed (`supervisor.ts`), ~18 lines added, 0 new dependencies
- No import changes (uses existing `fs`, `path`, `spawn`)
- No effect on macOS/Linux (`process.platform` guard)
- Falls back to original `provider` if resolution fails

## References
- [CVE-2024-27980](https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2) — Node.js security patch blocking `.cmd` spawn without `shell`
- #201 — same root cause (Pi-agent Windows spawn), different entry point

wsolarq11:
Fixes #348

`spawn("claude")` fails on Windows because npm puts only a `.cmd` shim in PATH, not the real `.exe`.  Node.js >=18.20.2 (CVE-2024-27980) blocks implicit `.cmd` execution without `shell: true`.  Using `shell: true` would reintroduce the injection vector -- the args carry `--append-system-prompt` content.

Instead, parse the npm-generated `.cmd` shim to find the real `.exe`, then spawn it directly.  No shell, no injection risk.

**Change:** `packages/cli/src/commands/channel/supervisor.ts`
- +`resolveProviderPath()` utility 
- 2 lines modified at the spawn call

- 0 import changes, 0 new dependencies
- No effect on macOS/Linux (`process.platform` guard)
- Fixes both `claude` and `codex` adapters
- Falls back to original `provider` if resolution fails

### DIFF
--- a/packages/cli/src/commands/channel/supervisor.ts
+++ b/packages/cli/src/commands/channel/supervisor.ts
@@ -76,6 +76,34 @@ type Child = ChildProcessByStdio<Writable, Readable, Readable>;
 const SHUTDOWN_GRACE_MS = 3000;
 
 /**
+ * On Windows, npm installs CLI tools as `.cmd` shims where the real `.exe`
+ * lives in `node_modules/`.  Node.js ≥18.20.2 (CVE-2024-27980) blocks
+ * implicit `.cmd` execution via `spawn()` without `shell: true`.  Resolve
+ * the real `.exe` path so we can spawn it directly — no shell, no injection.
+ */
+function resolveProviderPath(provider: string): string {
+  if (process.platform !== "win32") return provider;
+  try {
+    const cmdName = `${provider}.cmd`;
+    const dirs = (process.env.PATH ?? "").split(path.delimiter);
+    for (const dir of dirs) {
+      const cmdFile = path.join(dir, cmdName);
+      if (!fs.existsSync(cmdFile)) continue;
+      const content = fs.readFileSync(cmdFile, "utf8");
+      // npm-generated cmd shim format:  "%dp0%\node_modules\pkg\bin\name.exe" %*
+      const m = content.match(/"%dp0%\\(.+\.exe)"/i);
+      if (m) {
+        const exePath = path.join(dir, m[1]);
+        if (fs.existsSync(exePath)) return exePath;
+      }
+    }
+  } catch {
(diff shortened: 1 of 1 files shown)


taosu0216 (approved):
Reviewed the focused Windows spawn fix. CI and local verification passed; no blocking issues found. A Windows shim parsing regression test would still be useful as follow-up.

PR #433, opened 2026-07-15
feat(cli): add Grok Build platform support

jlxyfll:
## Summary

- add Grok Build as a first-class CLI platform, including `--grok`, platform registry/configuration, `.grok/skills`, flat commands, and Grok agent templates
- integrate Grok with the pull-based Class 2 workflow, Python `CLIAdapter`, platform detection, and `GROK_SESSION_ID` session-state lookup
- seed `implement.jsonl` and `check.jsonl` for Grok-only repositories and keep the published Python templates synchronized with Trellis's dogfood runtime copies
- add configurator, template, update, regression, and Python execution-level coverage

## Validation

- `pnpm test test/regression.test.ts -t '\[grok\]'` — 5 passed
- update workflow marker test — 1 passed
- related six-file suite — 511 passed; 2 environment-only failures because this checkout does not contain:
  - `marketplace/workflows/native/workflow.md`
  - `marketplace/workflows/tdd/workflow.md`
- full CLI suite — 1364 passed; the same 2 missing-marketplace-file failures, with no Grok regressions
- `pnpm typecheck` — passed
- `pnpm lint` — passed
- `pnpm lint:py` — passed with 0 errors (64 pre-existing unused-import warnings)
- `pnpm build` — passed
- `git diff --check` — passed

## Risk review

GitNexus comparison against the upstream base reported 21 tracked files, 37 symbols, 0 affected execution processes, and overall LOW risk. Runtime implementations with CRITICAL upstream blast radius were left unchanged and covered through execution-level regression tests instead.

## Summary by CodeRabbit

* **New Features**
  * Added **Grok Build** platform support end-to-end (init `--grok`, config generation, Grok command/skill/agent templates, and sub-agent workflow).
  * Added **Trae IDE** and **Oh My Pi (OMP)** platform recognition.
  * Updated Trellis workflow routing and sub-agent behavior for Grok.
* **Documentation**
  * Expanded platform file-map documentation with Grok Build, Trae, and OMP layout/capabilities.
* **Bug Fixes**
  * Improved Grok-enabled task context setup (JSONL seeding/tracking) and Grok session handling.
* **Tests**
  * Added/updated integration and template tests covering Grok and the new platforms.


taosu0216:
Pushed a follow-up commit (`09047aec`) fixing one real bug plus CodeRabbit's outstanding nitpicks, since they hadn't been addressed yet:

**Real bug — `GROK_SESSION_ID` never resolves in practice.** Tested against a real `grok` CLI (0.2.101) install and cross-checked docs.x.ai:
- `GROK_SESSION_ID` is a genuine Grok Build env var, but it is only injected into **hook script processes** (`docs.x.ai/build/features/hooks`), not into the bash-tool subprocess that actually runs `task.py`. Confirmed live: `grok -p "env | grep -i grok" --always-approve` only shows `GROK_AGENT=1` — no session id.
- Since Grok is installed as a "no hooks" pull-based platform here, this env var can never be populated where `active_task.py` reads it, so the code path was always dead in production.
- Fix: drop the fabricated `GROK_SESSION_ID` lookup and let Grok use the existing single-session fallback (`_resolve_single_session_fallback`) that Codex/Copilot/Gemini/Qoder sub-agents already rely on for the same reason (they don't inherit a parent session id either). This is strictly less code and matches established precedent rather than adding a new mechanism.
- Updated the regression test that asserted the old (never-true) behavior to assert the actual degraded-mode/fallback behavior instead.

**CodeRabbit nitpicks applied:**
- Consolidated the identical `omp`/`grok` branches in `cli_adapter.py`'s `get_commands_path` and `get_trellis_command_path`.
- Added `python`/`text` language identifiers to the flagged fenced code blocks.
- The `.omp/` section CodeRabbit flagged as missing from the bundled `platform-map.md` was already present as of the latest commit — false positive, no change needed there.
- Filled in the placeholder text in the dogfood session journal.

Also verified live against `grok inspect` that `.grok/agents/*.md` and `.grok/commands/*.md` are correctly discovered, and that `spawn_subagent`/`subagent_type` are the real dispatch tool/param names — those parts of the design are solid.

Validation: `pnpm test test/regression.test.ts -t '\[grok\]'` (5 passed), full CLI suite (1364 passed, 2 pre-existing marketplace-file failures unrelated to this change), `typecheck`/`lint`/`lint:py`/`build`/`git diff --check` all clean.

PR #439, opened 2026-07-16
fix(hooks): adapt SessionStart acknowledgment language

### LINKED ISSUE #412: SessionStart hook hard-codes Chinese in the first assistant response
## Summary

The generated SessionStart hook explicitly instructs the assistant to use Chinese in its first visible reply, regardless of the user's language or locale.

## Observed behavior

After starting a Claude Code session, running `/clear`, or after context compaction, the first assistant response includes a Chinese acknowledgment that the Trellis SessionStart context was loaded.

## Root cause

The shared SessionStart template contains a hard-coded language directive:

```python
FIRST_REPLY_NOTICE = """<first-reply-notice>
First visible reply: say once in Chinese that Trellis SessionStart context is loaded, then answer directly.
This notice is one-shot: do not repeat it after the first assistant reply in the same session.
</first-reply-notice>"""
```

Source:

`packages/cli/src/templates/shared-hooks/session-start.py`

The notice is then written into every SessionStart context payload:

```python
output.write(FIRST_REPLY_NOTICE)
```

For Claude Code, the generated `.claude/settings.json` runs this hook for `startup`, `clear`, and `compact`, so the behavior recurs for all three events.

## Expected behavior

Trellis should not force a specific natural language for all users. Preferably, the hook should not require a visible acknowledgment at all because it adds noise after startup, `/clear`, and compaction.

If an acknowledgment is considered necessary, it should either:

1. use the language of the current user/conversation;
2. use language-neutral wording; or
3. be configurable.

For example:

```text
First visible reply: briefly acknowledge in the user's current language that Trellis SessionStart context is loaded, then answer directly.
```

## Environment

- Trellis: `0.6.6`
- Platform: Claude Code
- OS: macOS

## Additional verification

The generated local hook's SHA-256 matched its entry in `.trellis/.template-hashes.json`, confirming that this was the unmodified Trellis-managed template rather than a project-local customization.

lifan-builds:
## Summary

- preserve Trellis's one-shot SessionStart proof-of-load acknowledgment
- choose its language from the user message that triggers the first visible reply
- fall back to an explicitly established project communication language, then `Trellis SessionStart ✓`
- prevent the acknowledgment from changing the language of the remaining response
- apply byte-equivalent wording across shared hooks, Codex, OpenCode, and Pi while keeping Copilot notice-free
- preserve existing payload shapes, context timing, OpenCode persistence/deduplication, and Pi memoization

## Why

The existing notice requires a fixed Chinese sentence. Because the notice is model-visible after startup, `/clear`, and compaction, it can steer an otherwise non-Chinese conversation into Chinese.

The acknowledgment was intentionally added as cross-platform proof that otherwise-hidden SessionStart context loaded, so this change keeps that observability instead of removing it. Language selection happens when the first visible reply is generated, when the triggering user message is normally available. No OS-locale or repository-content guessing is introduced.

## Validation

- full pre-commit suite: 333 core tests passed (1 skipped), 1,389 CLI tests passed
- focused SessionStart/OpenCode/Pi tests: 396 passed
- ESLint passed
- TypeScript typecheck passed
- basedpyright passed with 0 errors (64 pre-existing unused re-export warnings)
- all four live notices verified byte-equivalent
- fixed-Chinese wording and Copilot exclusion audits passed
- `git diff --check` passed

Fixes #412

## Summary by CodeRabbit

* **New Features**
  * Added adaptive first-reply acknowledgments based on the user’s request language, with project-language and language-neutral fallbacks.
  * Session-start context now includes additional startup information while preserving one-time acknowledgment behavior.
  * Acknowledgments remain consistent with the response language and are not repeated during the session.

* **Documentation**
  * Updated platform integration guidance and validation requirements for adaptive acknowledgments.

* **Tests**
  * Expanded coverage for acknowledgment content, ordering, persistence, and suppression of duplicate context injections.

### DIFF
--- a/.trellis/spec/cli/backend/platform-integration.md
+++ b/.trellis/spec/cli/backend/platform-integration.md
@@ -556,7 +556,7 @@ For Pi Agent:
 | ---------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | Session start                      | `session_start` extension event (notify-only; context-key is established but no prompt mutation)                                                                                                                                                                      |
 | Per-turn workflow-state breadcrumb | `before_agent_start.message` hidden custom message — persists cached `<workflow-state>` + `<session-overview>` from `getTurnCtx()` without showing it in the UI; skipped when identical to the last persisted runtime context (dedup)                                |
-| Startup context                    | `before_agent_start` builds compact SessionStart-equivalent context (`<first-reply-notice>`, `<session-overview>`, `<trellis-workflow>`) once per context key, memoizes it, and contributes the identical bytes to `systemPrompt` on every turn                       |
+| Startup context                    | `before_agent_start` builds compact SessionStart-equivalent context (`<session-context>`, adaptive `<first-reply-notice>`, `<session-overview>`, `<trellis-workflow>`, `<ready>`) once per context key, memoizes it, and contributes the identical bytes to `systemPrompt` on every turn |
 | Per-agent-invocation context       | `before_agent_start.systemPrompt` carries a per-context-key snapshot of task context (PRD + jsonl) taken on first use; later on-disk task changes are delivered through `before_agent_start.message` as `<trellis-task-context-update>` persisted messages           |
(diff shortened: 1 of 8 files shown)


taosu0216 (approved):
Reviewed in depth: notice wording verified byte-equivalent across shared/Codex/OpenCode/Pi templates, Copilot stays notice-free, full CLI suite 1389/1389 green, no payload-shape regressions. Thanks for the careful fix!

PR #468, opened 2026-07-23
feat(hooks): path-scoped on-demand spec injection

Zwdommy:
## What

When the agent touches a file, the specs that govern that file are injected right then — and **stay present as the session runs**: full text the first time, silent within a refresh window, then a compact ticket so rules never decay out of context; content changes and `/compact` re-teach in full.

## Commit map (6, chronological)

| Commit | What it is |
|---|---|
| `e7e56d8e` | **v1 mechanism** — `paths:` frontmatter on specs, PostToolUse hook, char budgets, pull mode (`get_context --mode spec`), 12 dogfood specs mapped |
| `3a6d5bcb` | **v2 ticket-refresh** (taosu's design) — stateless→ticket / first→full / sha-change→full / fixed-window silent / past-window ticket; identity ladder; user-global JSONL state |
| `b444f3ff` | Identity delegated to `resolve_context_key` (shared resolver, payload-first) |
| `49eccc9c` | **Review revision R1–R7** (response to taosu's review): budgets in characters (CJK no longer pays 3×), clock = real user turns + `compact_boundary` records (transcripts are append-only — verified), subagent derived transcript, single state file + flock, write-failure circuit breaker to stateless, scoped GC, deny-list globs, tolerant frontmatter, honest stateless ticket wording, one regex matcher, `spec_injection.tools` |
| `82e5bb48` | **R8** — UTF-8 truncation boundary fix back-ported to `inject-subagent-context.py` (+ same defect found & fixed in the Pi TS mirror), reviewer's exact repros as regression tests |
| `765e1cc3` | **Audit fix round F1–F13** — 41-agent adversarial self-audit found 32 confirmed issues; all fixed. Highlights: subagent clock actually ticks (beats = assistant messages there; user turns are structurally frozen at 1 — 58/58 real transcripts), injective identity sanitization, one canonical path normalization (+NFC, IGNORECASE on case-insensitive FS), flow-sequence `paths:`, `tools: []`, GC symlink containment, named-index reserve with `complete:false` re-teach (tickets never lie), duplication removed |

## Verification

- **1,641 tests green** (86 pin this feature; bite-proofed — reverting fixes turns suites red).
- **Real-state gates**: shipped scanner over 136 real transcripts on a dev machine (0 escaped exceptions); 58 real subagent transcripts prove the beat clock (2..244, median 44); all 9 real compact boundaries recognized; taosu's 9,353-char CJK spec injects whole; live dogfood E2E (first-touch FULL at 9,500 ceiling → silent → `complete:false` recorded).
- The hook ran organically on the authoring sessions themselves (dogfood): spec-context blocks injected, deduped, re-taught on sha change.

## Known bounds (documented in spec-injection.md / task design.md)

- Upstream `_sanitize_key` in the shared resolver can still fold exotic raw session ids before the hook sees them (fixing it means touching shared `active_task.py` — out of scope, documented).
- Specs larger than the per-event budget re-teach as truncated FULL each window (never a lying ticket); root cure is the rules-section authoring idea — separate question per review.
- Windows untested on real hardware; other platforms are wiring-only follow-ups; cross-platform shape lives in discussion #474.
- This base branch (`feat/v0.7-beta`) is outside the repo's CI trigger (`pull_request: branches: [main]`) — the local full gate runs the same commands.


taosu0216:
## Review

Strong work — this is well above the bar for hook code in this repo, and several things here are done properly:

- **`spec_match.py` glob engine** holds up under adversarial testing (23 cases: `**` across/zero segments, trailing `**` not matching the dir itself, segment-internal `**` degrading to `*`, literal `.` escaping, `?` strictness). `validate_glob`'s rejections match the docs.
- **Template ↔ dogfood copies are byte-identical** — verified on all three pairs. The only `settings.json` difference is the expected `{{PYTHON_CMD}}` resolution.
- **Uninstall is correct by construction** — `collectSharedHooks` keys off `SHARED_HOOKS_BY_PLATFORM`, so the new hook lands in `.template-hashes.json` with no uninstall changes. Ran the real `scrubHooksJson` against the new `settings.json`: the whole `PostToolUse` block is removed, user `env`/`enabledPlugins` preserved.
- **Fail-open discipline** — kill switches, bad stdin, non-dict payload, non-target tool, missing `file_path`, no `.trellis`, missing spec dir, import failure: all `exit 0` with empty stdout.
- **`spec-injection.md`** is one of the most honest design docs in this repo — it records rejected alternatives, the misfire-asymmetry principle, and why the ppid tier is deliberately unwired. The self-hosting `paths:` frontmatter is a nice touch.
- **Perf is fine** — 31 ms end-to-end on the 41-spec dogfood tree; a 61 MB transcript only pushes it to 49 ms.

That said, two things currently undercut the feature's core promise, and both are measurable rather than theoretical.

---

## 1. "Inject the full spec" is currently "inject the first 5%"

`max_spec_bytes` defaults to 8192. Of the **13 specs this PR gives `paths:` frontmatter, 11 exceed that cap**:

| bytes | spec |
|---:|---|
| 162,692 | `platform-integration.md` |
| 76,626 | `script-conventions.md` |
| 70,222 | `commands-channel.md` |
| 69,419 | `commands-mem.md` |
| 35,115 | `commands-update.md` |
| … | (6 more over 8,192) |

Editing `spec_match.py` injects `platform-integration.md` truncated to the first 8,192 bytes — **5.0% of the file, cut mid-markdown-table** — followed by a notice telling the model to go read a 162 KB file. That is the index-only mode the design doc explicitly rejects, reached by a different route.

CJK content pays twice: the budget is in **bytes** while the platform ceiling is in **characters**. A 9,353-character Chinese spec (comfortably inside the 10,000-char ceiling) injects only 2,846 characters — the whole `additionalContext` came to 2,944 characters, nowhere near the ceiling.

Suggestion: measure against the platform's own unit (characters), and raise the per-spec cap — or split large specs, but the cap shouldn't silently turn FULL into a 5% excerpt.

## 2. The refresh window measures transcript lines, which are not turns

`refresh_window_lines: 300` is documented as roughly 30 turns of breathing room. Measured against a real session transcript in this repo (`0c539091-….jsonl`):

```
total lines                1112
real user turns               9      ← type=="user", string content, not isMeta
everything else            1103      ← assistant 379, tool_result 136, attachment 353, snapshots, meta
lines per turn             ~124
300 lines                ≈ 2.4 turns
```

So the effective window is roughly an order of magnitude smaller than intended, and the ratio is unstable — tool-heavy turns run 100+ lines, conversational turns a handful.

Counting real turns is cheap and available: `type == "user"` **and** `message.content` is a string (not a `tool_result` array) **and** no `isMeta`.

## 3. The `/compact` branch never fires on Claude Code

`_within_window` treats a negative line delta as "transcript rewritten shorter, e.g. `/compact`", and both `spec-injection.md` and `spec-injection.integration.test.ts:483` freeze that as contract.

Claude Code transcripts are append-only. Compaction writes a `{"subtype":"compact_boundary"}` record and keeps appending. Scanned 2,917 real transcripts under `~/.claude/projects`: 51 contain a compact boundary, **none ever shrank**.

Net effect: after `/compact` the spec body is gone from context, the line clock is still climbing, so the hook reads "inside the window" and stays silent — exactly the moment the ticket model exists for. Detecting the `compact_boundary` record, or falling back to the wall clock, would close this.

## 4. Subagents measure the parent's clock — and their own transcript exists

`transcript_path` in a subagent's PostToolUse payload points at the **parent** session file. While a subagent works, the parent is idle, so the line count barely moves — a subagent can run 100+ turns and never leave the window. Subagents are precisely the long-horizon case this feature targets.

Claude Code does write a per-subagent transcript:

```
<project>/<parent-uuid>.jsonl                              ← parent
<project>/<parent-uuid>/subagents/agent-<agentId>.jsonl    ← subagent (every line isSidechain:true)
<project>/<parent-uuid>/subagents/agent-<agentId>.meta.json
```

(2,351 such files on this machine.) The hook already receives `agent_id` and uses it for state separation — the same value derives the subagent's own transcript path, which fixes the clock.

## 5. Per-pid state sharding isn't needed, and doesn't solve what it looks like it solves

The comment says sharding avoids concurrent-write conflicts without locking. Measured: 12 processes each appending **38,633 bytes in a single `os.write()`** (far past `PIPE_BUF`) to one `O_APPEND` file → 2,400 lines, zero corruption. Real records are 200–600 bytes, an order of magnitude safer. Appending short lines under `O_APPEND` is what every log file does.

The cost is real, though: each hook run is a fresh process, so the pid differs every time — sharding becomes **one file per emission**, and `load_state` globs and merges all of them on every event, growing linearly with session length.

And the duplicate-injection race is untouched by sharding: concurrent hooks all *read* state before any of them *writes*, so they all conclude "first touch". Measured on 6 concurrent tool calls against one governing spec: 2–5 processes each emitted a full `<spec-context>`, 17–43 KB in a single turn.

One file per session with `O_APPEND` is simpler and equally safe; a lock between read and write would additionally close the duplicate-injection race.

## 6. Fail-closed gaps

- **State dir unwritable** → `append_records` warns and returns, but `stateless` stays `False`, so every subsequent event reads empty state, decides "first touch", and re-emits the full body. Reproduced: same payload three times in a row, 8,986 bytes each. The stateless ticket-only mode already exists — this path should fall into it.
- **`_maybe_gc` deletes by extension only** — `rglob("*.jsonl")` with an mtime check, no shard-name validation and no depth limit, on a `base_dir` taken straight from `TRELLIS_SPEC_STATE_DIR` with no validation. Pointing that variable at a non-dedicated directory silently deletes user `.jsonl` files.

## 7. Smaller, verified

- **Glob charset whitelist** `^[A-Za-z0-9_./*?-]+$` rejects `packages/@scope/**` (npm workspaces), `app/[slug]/page.tsx` and `app/(marketing)/**` (Next.js App Router), and non-ASCII directory names. Rejection only warns on stderr — and a PostToolUse hook's stderr isn't surfaced in Claude Code, so a mistyped glob fails silently forever.
- **Stateless tier ticket text is untrue.** `decide()` returns `"ticket"` unconditionally when stateless, but `_ticket_block()` opens with "You were shown this spec earlier in this session and its content is unchanged." In that tier the body was never injected — and the wording invites the model to skip reading it, which is the exact failure mode the feature targets.
- **Frontmatter parsing is all-or-nothing.** Any line that isn't `key:` or `- item` raises, and the caller drops the whole spec including a perfectly valid `paths:`. A block scalar (`description: >`) — the style Claude skill `SKILL.md` files use — takes the spec out entirely, with only an stderr warning.
- **Budget accounting misses two things**: the `(+N more …)` summary line is appended after the budget loop without a check, and the `"\n\n"` separators between blocks are never counted, so `max_total_bytes` isn't a hard ceiling.
- **`stdin` isn't reconfigured on Windows**, only `stdout` — while this PR's own `script-conventions.md` (which it puts under `paths: shared-hooks/**`) has a section titled "CRITICAL: Windows stdio Encoding".
- **`settings.json` repeats the same PostToolUse entry four times**; Claude matchers take regex (`cursor/hooks.json` already uses `"Task|Subagent"`), so `"Read|Edit|Write|MultiEdit"` would do. Also worth making the trigger tool set configurable — "spawn a Python process on every Read" is likely the first thing someone will want to turn off.
- **Wiring is not frozen.** Adding the `PostToolUse` entry means existing users hit a whole-file conflict on `settings.json` during `update`, where the default choice is skip — leaving the script installed but unwired, with no notice.

## Pre-existing bug found along the way

`truncate_utf8` in `inject-subagent-context.py` (the version this PR's copy was derived from) splits multi-byte characters when the cut lands exactly on a character boundary:

```
"你好世界" capped at 6 bytes  →  old: 你␦   new (this PR): 你好
"a\xc3\xa9b" capped at 3      →  old: a␦    new (this PR): aé
```

Not introduced here — this PR's copy fixes it — but the original is still live, and our specs are Chinese, so subagent context truncation corrupts the last character today. Worth back-porting the fix rather than leaving two diverging copies.

---

## On structure

The seams are in the right places: matching lives in its own module, `decide()`/`_within_window()` are pure, identity genuinely delegates to `resolve_context_key`, envelopes have renderers. Swapping a decision policy or a glob rule really is a one-function change.

Two things limit that, though:

1. **`build_payload` does seven jobs** (read, hash, decide, truncate, render, budget, record). It's also why the hot path can't avoid the IO — `read_bytes()` + `sha256` happen *before* `decide()`, so even a `"silent"` outcome pays for a full file read.
2. **No test calls the pure functions.** 19 of 33 cases spawn the script and assert on verbatim output strings, so a wording change breaks ~10 tests while a decision-engine refactor has no fine-grained net. `runSpecProbe` already exists and does exactly the right thing for `spec_match` — it just can't reach the hook, because the filename has hyphens and there's no importable core module.

Extracting the pure logic (decide / window / truncate / render / budget) into `common/spec_inject.py` would fix both, and would let `truncate_utf8` exist exactly once in the repo.

---

Happy to go deeper on any of these — most were reproduced against the branch with real `python3`, and I can share the fixtures. Suggested order: the fail-closed gaps (6) and the GC delete scope first, then the clock (2/3/4), then the budget (1).

🤖 Findings reproduced with [Claude Code](https://claude.com/claude-code); each was independently re-verified before inclusion.

taosu0216:
## A simplification worth considering before patching the state layer

Posting this now rather than later: several items in my previous comment are patches to the state/clock design, and if this direction holds, most of them stop existing rather than needing fixes. Better to raise it before you spend a day on the sharding and GC.

Everything in the previous comment stands as-is for the current design — this is orthogonal, and you may well have considered and rejected it.

### The observation

The refresh state we're maintaining externally is already recorded in the transcript, in a form that's strictly more accurate than our own bookkeeping. Three record types, all verified on real transcripts under `~/.claude/projects`:

**1. Our own past injections are recorded.**
```json
{"type":"attachment","attachment":{"type":"hook_additional_context","content":["<workflow-state>…"]}}
```
(Found by grepping for this repo's own `UserPromptSubmit` hook output — 16 occurrences in one session file.)

**2. Compaction is recorded, with exactly what survived.**
```json
{"type":"system","subtype":"compact_boundary",
 "compactMetadata":{"trigger":"manual","preTokens":484016,"postTokens":10315,
                    "cumulativeDroppedTokens":473701,
                    "preservedSegment":{"headUuid":"94d2fde2-…"}}}
```
Pre-compaction records stay in the file (one sample: boundary at line 2173 of 2331), so "what the model currently sees" is exactly *the tail after the last boundary*.

**3. Context size in tokens is recorded per assistant turn.**
```json
"usage":{"input_tokens":21047,"cache_creation_input_tokens":19357,"cache_read_input_tokens":18365,…}
```
379 such records in one session. Summing the input side at any point gives the context size at that moment, so *context growth since an injection* is directly measurable — not estimated.

### What that enables

Instead of "did I inject this, and how many lines/seconds ago", scan the transcript backward from the tail:

- **Hit a `compact_boundary` before finding the marker** → the injection was compacted away. Re-inject **FULL** — the text is genuinely gone, a ticket would be wrong here.
- **Find the marker** → compare context size then vs now (record 3) → past the threshold, refresh.
- **Nothing** → first touch.

### What this removes

| Currently needed | With transcript-as-state |
|---|---|
| Identity ladder (payload → env → fallbacks) | Not needed — the transcript path *is* the session handle |
| State files, per-pid shards, merge-on-read | Not needed |
| GC + the `rglob("*.jsonl")` delete hazard | Not needed |
| Negative-line-delta as a compaction proxy | Not a proxy — the boundary record is right there |
| `refresh_window_lines` / `refresh_window_seconds` | Replaced by the causal quantity: context tokens accumulated |
| Subagent measuring the parent's clock | Subagent has its own transcript with its own markers |

### On identity specifically

This doesn't dodge the identity question — it replaces it with an easier one. The hook is *handed* `transcript_path` in the payload; it doesn't have to resolve an identity and then derive a filename. Your code already reads that field for the line clock, so the dependency exists today.

Two properties that the current ladder can't offer:

- **Collision is impossible by construction.** The whole ladder exists to prevent two live sessions collapsing onto one identity (→ missed injection, the unacceptable direction). Two sessions cannot share a transcript file. Note that `_sanitize` currently maps disallowed characters to `_`, so two distinct upstream keys *can* fold onto one identity — a path can't.
- **It's self-validating.** Records carry `sessionId`, and the filename is that same UUID (verified). Read the wrong file and you can tell.

Fallbacks, in order:
1. `transcript_path` from the payload.
2. Derive it: `~/.claude/projects/<cwd with `/` and `_` → `-`>/<session_id>.jsonl` (verified convention; both inputs are in the payload).
3. Neither available → stateless ticket-only, same as the current T4 tier.

For subagents, derive the child path from the parent's: `<project>/<parent-uuid>/subagents/agent-<agent_id>.jsonl` — fixed convention, 2,351 such files on this machine, and `agent_id` is already in the payload (you're using it for state separation today).

### Honest caveats

- **The same-turn concurrency race is unchanged.** Sibling injections aren't in the transcript yet either. Bounded, not solved.
- **This is Claude-Code-shaped.** Other platforms' transcript formats differ, so it needs a per-platform reader — though this iteration is Claude-only anyway, and the ticket-only degradation covers the rest.
- **Cost is comparable, not free.** You're already reading the transcript for line counting; scanning the tail for markers and one `usage` field is the same order, and can be bounded the same way (read tail-first, stop at the last boundary).
- **Marker format becomes a compatibility surface.** If the injected block's shape changes, older markers stop matching → re-inject. Safe direction, but worth pinning the marker as a stable contract separate from the human-readable wording.

### Separate thought on the budget

Possibly the 8,192-byte problem isn't a sizing problem. We're pushing 162 KB of human-oriented prose through a 10,000-character channel; at edit time the model needs the normative constraints, not the rationale and examples. If specs carried an explicit rules section (or the injector extracted the imperative lines), FULL would be small by construction — no truncation, no FULL-vs-TICKET distinction, and refreshing would be cheap enough that the window matters less.

Related: a ticket is arguably the more expensive option in exactly the case it's meant for. If the model still remembers, it's wasted tokens; if it forgot, it costs a round trip *and* the full spec enters context anyway.

That one is a bigger change than the rest of this and touches spec authoring conventions, so treat it as a separate question.

---

Happy to prototype the backward-scan reader against the branch if it's useful — the record shapes above are all verified, and I can share the probe scripts.

🤖 Verified with [Claude Code](https://claude.com/claude-code) against real transcripts; record structures quoted verbatim.

taosu0216:
Moved the cross-platform half of my earlier comment to a discussion so it doesn't get buried when this merges: **#474**.

Keeping this PR scoped to the Claude Code implementation is right. What the discussion covers instead is the shape that lets platform #2 be a descriptor rather than a second injector — a narrow provider contract (`ContextView`: ordered entries with compaction already applied) under a single shared decision / budget / envelope path, with the three JSONL platforms sharing one reader plus a per-platform dialect, and the two in-process ones (Pi, OpenCode) gathering the view in their own runtime but delegating the decision.

For this PR specifically, the findings that stand on their own regardless of that direction:

- **The `/compact` branch never fires on Claude Code** — transcripts are append-only (compaction writes `compact_boundary` and keeps appending; 51 of 2,917 scanned transcripts contain a boundary, none shrank). `compact_boundary` is the signal to key off instead.
- **`refresh_window_lines: 300` is ≈ 2.4 turns, not ≈ 30** — measured 1,112 lines / 9 real user turns in a session here. Counting `type == "user"` with string content and no `isMeta` gives turns directly.
- **Subagents measure the parent's clock** — and they have their own transcript at `<project>/<parent-uuid>/subagents/agent-<agent_id>.jsonl`, derivable from the `agent_id` already in the payload.
- **State-dir write failure has no circuit breaker** — `stateless` stays `False`, so every subsequent event re-decides "first touch" and re-emits the full body (reproduced: 8,986 bytes, three runs in a row). The stateless ticket-only mode already exists; this path should fall into it.
- **`_maybe_gc` deletes by extension only** — `rglob("*.jsonl")` + mtime, no shard-name validation, no depth limit, on an unvalidated `TRELLIS_SPEC_STATE_DIR`.

The rest of my first comment (budget sizing, glob charset, stateless ticket wording, frontmatter all-or-nothing, stdin encoding) is unaffected either way.

PR #510, opened 2026-07-31
fix(codex): read platform skills from codex/skills/ not codex-skills/

jlxyfll:
## Problem

`getAllCodexSkills()` in `packages/cli/src/templates/codex/index.ts` lists `codex-skills` and reads `codex-skills/${name}/SKILL.md`, but the template directory was moved to `codex/skills/` (the 13 skills now live there). Since the move, the function always returns `[]`, so `trellis update` never writes the Codex platform skills to `.codex/skills/` even though it reports `Already up to date!`.

## Fix

Point both the directory listing and the read path at `skills/` (the actual template dir). Updated the doc comment and the regression test (it previously asserted the empty return, which baked in the bug).

## Verification

- `pnpm run typecheck` ✅
- `eslint` on changed files ✅
- `vitest run test/templates/codex.test.ts test/configurators/codex.test.ts` → 35/35 pass ✅
- Full suite: only the pre-existing `platforms.integration.test.ts` failures (3, unrelated, fail on clean main too)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated Codex skill discovery to use the standard `skills/` directory.
  * Codex-specific skills are now correctly installed and recognized under `.codex/skills/`.
  * Updated skill validation to reflect the correct names and metadata.

### DIFF
--- a/packages/cli/src/templates/codex/index.ts
+++ b/packages/cli/src/templates/codex/index.ts
@@ -8,8 +8,7 @@
  * Directory structure:
  *   codex/
  *   ├── agents/         # Project-scoped Codex custom agents (.toml)
- *   ├── codex-skills/   # Codex-specific skills → .codex/skills/
- *   ├── skills/         # Shared skills → .agents/skills/
+ *   ├── skills/         # Codex-specific skills → .codex/skills/
  *   └── config.toml     # Project-scoped Codex config
  */
 
@@ -82,8 +81,8 @@ export function getAllAgents(): AgentTemplate[] {
 export function getAllCodexSkills(): SkillTemplate[] {
   const skills: SkillTemplate[] = [];
 
-  for (const name of listDirectories("codex-skills")) {
-    const content = readTemplate(`codex-skills/${name}/SKILL.md`);
+  for (const name of listDirectories("skills")) {
+    const content = readTemplate(`skills/${name}/SKILL.md`);
     skills.push({ name, content });
   }
 
--- a/packages/cli/test/templates/codex.test.ts
+++ b/packages/cli/test/templates/codex.test.ts
@@ -92,9 +92,13 @@ describe("codex native sub-agent hooks", () => {
 });
 
 describe("codex getAllCodexSkills (platform-specific)", () => {
-  it("returns empty after parallel removal", () => {
+  it("returns the platform skills installed to .codex/skills/", () => {
     const skills = getAllCodexSkills();
-    expect(skills).toEqual([]);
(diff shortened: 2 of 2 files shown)


taosu0216:
Thanks for digging into this — the symptom you describe is real: `trellis update` never writes anything to `.codex/skills/`. But the empty return is intentional, not a stale path, so pointing the reader at `skills/` would reintroduce a duplicate rather than fix a gap.

Where the skills actually come from now:

- Shared workflow skills are sourced from `templates/common/` and written to `.agents/skills/` as `trellis-*` (`src/configurators/codex.ts:158`, via `resolveAllAsSkillsNeutral` / `resolveBundledSkills`).
- Codex reads both `.agents/skills/` and `.codex/skills/`, so those skills are already installed and reachable.
- `src/templates/codex/skills/` is a leftover copy from before that consolidation. Its 13 entries are the same skills as `templates/common/skills/`, just pre-wrapped with SKILL.md frontmatter. Repointing `getAllCodexSkills()` at it would install a second copy of every skill under `.codex/skills/`.

That is also what the existing assertion `"returns empty after parallel removal"` in `test/templates/codex.test.ts` is guarding — it records the removal as deliberate.

The real cleanup is to delete `packages/cli/src/templates/codex/skills/` and fix the now-inaccurate directory comment at `src/templates/codex/index.ts:11`, which still documents a `codex-skills/` directory that no longer exists. That stale comment is very likely what sent you down this path — sorry for the wasted trip.

Closing this one, but the report was useful. If you'd like to send the deletion PR instead, that would be welcome.

taosu0216:
Follow-up: the cleanup is done in 82fc41a2 on `main` — `src/templates/codex/skills/` is gone, along with the stale directory comment at `templates/codex/index.ts:11` that still described a `codex-skills/` directory. That comment is almost certainly what pointed you at the wrong fix, so thanks for surfacing it.

One correction to my earlier comment: I said the 13 entries were the same skills as `templates/common/skills/`. That was only true for 7 of them. The other 6 — `check-cross-layer`, `create-command`, `improve-ut`, `integrate-skill`, `onboard`, `record-session` — exist nowhere in the common pipeline (Copilot carries its own `.prompt.md` versions). They were not reaching Codex users either way, since the loader has returned `[]` since the `parallel` removal, so deleting them changes nothing at install time. But "duplicate copy" was not an accurate description of the whole directory.

The conclusion is unchanged: the directory was never installed, and wiring it up would have double-installed the 7 overlapping skills.

PR #524, opened 2026-08-05
fix(opencode): persist synthetic context outside user messages

kakamisamas:
## Summary

- persist SessionStart and workflow-state context as separate synthetic OpenCode text parts
- preserve ordinary user text, file parts, metadata, identity, and relative order
- assign deterministic complete part identities so first-use order matches stored replay order
- keep template and dogfood copies byte-identical and document the persisted-part contract

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm typecheck`
- `pnpm --filter @mindfoldhq/trellis lint:py` (0 errors; existing unused-import warnings only)
- `pnpm lint`
- `pnpm build`
- `pnpm test` (Core: 333 passed, 1 skipped; CLI: 1645 passed; OpenCode: 70 passed)
- `git diff --check`
- template/dogfood byte comparison and `node --check` for all changed JavaScript files
- GitNexus change analysis: low risk, 0 affected processes

## Risk

The identity helper intentionally accepts the current persisted OpenCode `prt_` ID format. Unsupported or incomplete identities fail closed: plugins log the error and preserve all ordinary input parts unchanged. Regression coverage includes both plugin orders, attachment-only turns, duplicate refusal, persisted-history dedupe, compaction, and ID-sorted replay.

## Rollback

Revert commit `ac66e207`; this restores the previous in-place context injection behavior without a data migration.

## Summary by CodeRabbit

* **New Features**
  * Added persisted synthetic context entries for session-start and workflow-state information.
  * Preserved original prompts while keeping injected context separate.
  * Added deterministic ordering and deduplication for injected context.

* **Bug Fixes**
  * Prevented context injection from modifying user-authored text.
  * Improved handling of invalid or missing message information.

* **Tests**
  * Expanded coverage for persistence, ordering, validation, skip conditions, replay behavior, and prompt preservation.

### DIFF
--- a/.opencode/lib/context-visibility.js
+++ b/.opencode/lib/context-visibility.js
@@ -1,0 +1,82 @@
+import { createHash } from "node:crypto"
+
+const PART_ID_PATTERN = /^prt_([0-9a-f]{12})[0-9A-Za-z]{14}$/
+const CONTEXT_PART_KINDS = {
+  sessionStart: { offset: 2n, slot: "0" },
+  workflowState: { offset: 1n, slot: "1" },
+}
+
+/**
+ * Return the first ordinary user-authored text part.
+ *
+ * Trellis plugins can run in either order, so callers must not mistake a
+ * synthetic context part inserted by another plugin for the user's prompt.
+ */
+export function findUserTextPart(parts) {
+  if (!Array.isArray(parts)) return undefined
+  return parts.find(
+    part => part?.type === "text" && part.synthetic !== true && part.text !== undefined,
+  )
+}
+
+function findIdentitySourcePart(parts) {
+  return parts
+    .filter(
+      part =>
+        part?.synthetic !== true &&
+        typeof part?.id === "string" &&
+        typeof part?.sessionID === "string" &&
+        typeof part?.messageID === "string" &&
+        PART_ID_PATTERN.test(part.id),
+    )
+    .sort((left, right) => (left.id < right.id ? -1 : left.id > right.id ? 1 : 0))[0]
+}
+
+function contextPartId(source, kind) {
+  const definition = CONTEXT_PART_KINDS[kind]
+  if (!definition) throw new TypeError(`unknown context part kind: ${kind}`)
+
+  const match = PART_ID_PATTERN.exec(source.id)
+  if (!match) throw new TypeError(`unsupported OpenCode part ID: ${source.id}`)
+  const sourceOrdinal = BigInt(`0x${match[1]}`)
+  if (sourceOrdinal <= definition.offset) {
+    throw new TypeError(`unsupported OpenCode part ID ordinal: ${source.id}`)
+  }
+  const ordinal = sourceOrdinal - definition.offset
+  const suffix = createHash("sha256")
+    .update(`${source.messageID}\0${kind}`)
+    .digest("hex")
+    .slice(0, 13)
+  return `prt_${ordinal.toString(16).padStart(12, "0")}${definition.slot}${suffix}`
+}
+
+/**
+ * Persist machine-authored context ahead of the ordinary user parts without
+ * mutating any existing part. The generated identity sorts before the source
+ * user part in the same order during both the first request and stored replay.
+ */
+export function insertSyntheticTextPart(parts, text, kind) {
+  if (!Array.isArray(parts)) throw new TypeError("parts must be an array")
+  if (typeof text !== "string") throw new TypeError("text must be a string")
+  if (!CONTEXT_PART_KINDS[kind]) throw new TypeError(`unknown context part kind: ${kind}`)
+
+  const source = findIdentitySourcePart(parts)
+  if (!source) throw new TypeError("no ordinary OpenCode part with a persisted identity")
+
+  const id = contextPartId(source, kind)
+  if (parts.some(part => part?.id === id)) {
+    throw new TypeError(`duplicate synthetic context part: ${kind}`)
+  }
+  const part = {
+    id,
+    sessionID: source.sessionID,
+    messageID: source.messageID,
+    type: "text",
+    text,
+    synthetic: true,
+  }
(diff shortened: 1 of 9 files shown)


taosu0216:
Reviewed this in depth. The premise holds and I verified it against OpenCode 1.17.18's actual implementation rather than against the description — extracted the embedded JS from the installed binary and checked a live `opencode.db`:

- `SessionPrompt.createUserMessage` stamps `id`/`messageID`/`sessionID` onto every part **before** triggering `chat.message`, then iterates the same mutated array calling `updatePart`. So in-place `splice`/`push` is the correct contract, and the old ID-less `parts.unshift(...)` was indeed handing OpenCode a part it had to invent an ID for.
- The read path is `orderBy(part.message_id, part.id)`, so lexicographic part ID really is replay order. The central claim of the design is true.
- `synthetic: true` is a real schema field, and the model conversion filters on `!part.ignored` while the UI filters on `!synthetic && !ignored`. That is exactly the right split — it also fixes a real annoyance where reverting a message restored the user's prompt with ~8KB of Trellis context glued to the front.
- The bug is currently shipping: a part in today's dogfood session is a single 8470-byte text part containing `<session-context>…\n\n---\n\n<workflow-state>…` followed by the user's actual sentence.

It also rebases onto current `main` with no conflicts, and the full suite is green there (1658 passed / 72 files), with the `.opencode/` dogfood copies byte-identical to the templates. Thanks for keeping those in sync — that seam has bitten us before.

Three things I'd like before merge. Two are missing tests that let real mutations survive; I ran the mutations rather than eyeballing coverage.

**1. The "identity source must be the minimum part" invariant is undefended.**

I inverted the comparator in `findIdentitySourcePart` (`context-visibility.js:26-37`) so it picks the largest ordinary part instead of the smallest. **Zero tests failed** — 489 passed unchanged. No fixture anywhere puts two ordinary parts in one array; the attachment tests use attachment-only messages, so "typed text plus a dragged-in file" is never exercised.

Why it matters, with real ID shapes from the live DB — text `prt_b9634c88c001…` and file `prt_b9634c88c002…`, same millisecond, counters 001/002:

- min-source (shipped): synthetic ordinals `…bfff` / `…c000`, both below both real parts, correctly borrowing across the millisecond boundary.
- max-source: workflow-state gets ordinal `…c001`, an *identical* 12-hex prefix to the user's text part. Ordering then falls to the 14-char tail, and a real tail beginning with `0` (1 in 62) puts the breadcrumb after the user's prompt.

Min-selection is what makes the scheme safe. Could you add a case with a text part **and** a file part asserting both synthetic IDs sort below both ordinary parts?

**2. The ID-format strictness is also untested, and drift is not hypothetical.**

Loosening `PART_ID_PATTERN` (`context-visibility.js:3`) from `[0-9A-Za-z]{14}$` to `[0-9A-Za-z]*$` likewise killed **zero tests**. The existing "missing identity" case only covers a part with no `id` at all.

In the live DB, 2435/2440 IDs match the current pattern and **5 use an older 19-char format** (`prt_19b9634d8e5924s6zx6`) — OpenCode has already changed this once. When it changes again, injection stops entirely on OpenCode. User parts are preserved, so the fail-closed claim is true, but the only signal is an append to the hardcoded `/tmp/trellis-plugin-debug.log`, which is swallowed on Windows.

Worth noting the pattern is stricter than the algorithm needs: only the leading 12 hex participate in ordering, so a tail-length change alone would be harmless to correctness yet currently kills injection. A test using that real legacy ID, asserting user parts survive untouched, would pin the intended behavior either way.

**3. The `session.compacted` handler is dead code, and the rewritten test now certifies it.**

`session-start.js:23-37` clears the in-memory flag so context can be re-injected after compaction. It cannot work: `client.session.messages` returns the whole session (`filterCompacted` applies only when building model messages), so `hasPersistedInjectedContext` still finds the pre-compaction `metadata.trellis.sessionStart` marker and re-suppresses immediately. On overflow-triggered compaction OpenCode also copies the previous user message's non-compaction parts forward into a fresh message, carrying the marker with them.

The new test fires `session.compacted` and then asserts the second message gets no injection — which documents the handler doing nothing, while the comment above it says "Cleared processed flag after compaction". Either delete the handler or make the dedupe compaction-aware; a test that makes dead code look deliberate is worse than no test.

(For what it's worth, I checked the two failure modes I was worried about and both are fine: double-injection after compaction cannot happen, and permanent suppression from a false-positive dedupe is not realistic, since a `session.messages` failure returns `false` and so fails toward re-injecting.)

**Optional, cheap:** `contextCollector.markProcessed(sessionID)` sits at `session-start.js:84`, after the throwing insert. If the ID format is ever unsupported, every subsequent message re-runs a full `session.messages` fetch plus a full `buildSessionContext` filesystem walk and throws again, for the rest of the session. Moving the mark above the insert, or adding a per-session "identity unsupported" latch, bounds that. This retry loop is new — the pre-PR path could not fail.

A few things I checked and found clean, so you don't need to re-litigate them: `inject-subagent-context.js` is genuinely unaffected (it only registers `tool.execute.before`, never reads `output.parts`, and neither `lib/trellis-context.js` nor `lib/session-utils.js` is in the diff — the `export TRELLIS_CONTEXT_ID=…` bash prefixing is untouched); ID collision analysis holds up given `part.id` is a global primary key; and `lib/context-visibility.js` earns its place, since `insertSyntheticTextPart` has two call sites in two separately-loaded plugin modules and `CONTEXT_PART_KINDS` is the shared contract that makes them order correctly regardless of load order.

Test quality is otherwise strong — I ran 12 mutations and 10 were killed, several by many tests at once, and the `structuredClone` deep-comparison baselines are doing real work. The two survivors above are the gaps.

One thing I could not determine: whether the desktop/web client renders synthetic user text parts inside the message bubble. I confirmed they are filtered out of prompt reconstruction, timeline, fork, copy and revert. Cosmetic either way, and no worse than today's inline concatenation.

PR #541, opened 2026-08-12
fix(omp): refresh task context after file changes

izumi0uu:
## 摘要

修复 OMP 会话启动后任务上下文不随文件修改刷新的问题。

## 背景

OMP 原先只在 `session_start` 时读取当前 task 的 `prd.md`、`info.md` 以及 manifest 引用的文件，并通过持久化的 `trellis-task-context` 消息注入。会话继续后，文件即使已经更新，模型仍可能看到启动时的旧快照，与工具读取到的新文件冲突。

## 实现

- 为 task context 建立输入文件签名，包含 `mtime`、`ctime` 和大小。
- 未发生变化时复用当前会话内的上下文缓存，不重复读取和投影。
- 发现 task artifact、JSONL manifest 或引用文件变化时，重新构建 task context。
- 在 OMP 的 `context` 投影阶段替换旧的 `trellis-task-context`，避免把新旧正文同时发送给模型。
- 支持启动时缺失、会话中后来创建的 manifest 引用文件。
- 同步更新 OMP 项目扩展与 CLI 安装模板。

## 测试

- OMP 模板定向测试：`14/14` 通过。
- Core 测试：`344` 通过，`1` 个跳过项。
- CLI 完整测试：`1672/1672` 通过。
- lint、typecheck、build 和 `git diff --check` 已通过。

## 关联

关联 Issue #533。

## Summary by CodeRabbit

* **Bug Fixes**
  * Task context now refreshes automatically when referenced files change or new files are added.
  * Stale or duplicate task-context messages are replaced or removed.
  * Updated context is consistently applied across main sessions and sub-agent sessions.
  * Workflow state remains available after context compaction.
  * Unchanged context is no longer unnecessarily reinjected, improving session efficiency.
  * Task context is removed cleanly when the active task is no longer available.

### DIFF
--- a/.omp/extensions/trellis/index.ts
+++ b/.omp/extensions/trellis/index.ts
@@ -1,5 +1,5 @@
 import type { ExtensionAPI } from "@oh-my-pi/pi-coding-agent";
-import { existsSync, lstatSync, readFileSync, readdirSync, realpathSync } from "node:fs";
+import { existsSync, lstatSync, readFileSync, readdirSync, realpathSync, statSync } from "node:fs";
 import { join, dirname, isAbsolute, relative, resolve } from "node:path";
 import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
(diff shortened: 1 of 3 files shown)


taosu0216:
Merged #540 first, so this one needs a rebase before it can go in — sorry for the wait on all three of these.

## Why the order

All three OMP PRs branch from the same base and rewrite the same loop in `buildTaskContext`, so they conflict with each other rather than stacking:

```
540 × 541  → conflicts in omp.test.ts
540 × 543  → conflicts in all three files
541 × 543  → conflicts in all three files
```

#540 was the most self-contained (+90/-0, dedup by realpath) so it went first. That leaves this one and #543, in that order.

## What we checked here

The signature approach (mtime + ctime + size over prd/info/the jsonl files/every manifest-referenced file) is the right shape for this — the bug is real, and a session that edits a referenced file today keeps seeing the startup snapshot while its own tools read the new content.

We also saw the CodeRabbit point about the first version overwriting a file with same-length content, where a coarse filesystem clock leaves `mtimeMs`/`ctimeMs`/`size` unchanged and the test passes for the wrong reason. Your `ff137c9` fix (different-length content) is the right one — that class of test is worth being fussy about, since it fails open.

## What we need

Rebase onto `main` (now at `a8a50a5e`). The conflict with #540 is confined to `omp.test.ts`; the dedup `Set` it added lives in the same loop your signature cache wraps, so please keep both behaviours rather than taking one side wholesale — a plain "ours"/"theirs" resolution will silently drop one of the two fixes.

Ping here once it is rebased and I will re-run the suite and merge.

izumi0uu:
Rebased this PR onto the latest `main` (`a8a50a5e`).

I resolved the OMP test conflict by preserving both #540's realpath-based deduplication coverage and this PR's task-context refresh coverage.

Local verification passed:
- CLI: 1,457 tests
- Core: 344 passed, 1 skipped
- Typecheck
- ESLint
- Build
- OMP source/template parity

The PR is ready for another review. Thanks!

PR #545, opened 2026-08-13
fix(hooks): surface unreadable active task records (#544)

### LINKED ISSUE #544: inject-workflow-state.py: docstring promises a silent path for a malformed task.json, but it is reported as 'no active task'
### Summary

The module docstring of `inject-workflow-state.py` lists a silent no-output path
that the code does not have. A malformed or status-less `task.json` is
documented as producing no output, but it is actually reported to the model as
"no active task" — the opposite of the truth, since a task *is* active and its
record is simply unreadable.

### Where

`packages/cli/src/templates/shared-hooks/inject-workflow-state.py` on `main`
(also vendored to `.claude/hooks/` and `.codex/hooks/`).

The docstring, lines 30–32:

```
Silent exit 0 cases (no output):
  - No .trellis/ directory found (not a Trellis project)
  - task.json malformed or missing status
```

`get_active_task` returns `None` for exactly that case — unparseable JSON or an
unusable `status` (lines 175–176 and 180–181):

```python
    try:
        data = json.loads(task_json.read_text(encoding="utf-8"))
    except (json.JSONDecodeError, OSError):
        return None
    ...
    if not isinstance(status, str) or not status:
        return None
```

But `main` treats `None` as "no active task" and emits a breadcrumb anyway
(line 427):

```python
    task = get_active_task(root, data)
    if task is None:
        # No active task — still emit a breadcrumb nudging AI toward
        # trellis-brainstorm + task.py create when user describes real work.
```

Only the missing-`.trellis/` branch above it actually returns early without
output. The second bullet has no corresponding code path.

### Why it matters

The two states need opposite responses, and the hook collapses them into one:

- **Genuinely no active task** — the `no_task` breadcrumb is right. It nudges
  toward `trellis-brainstorm` and `task.py create`.
- **Active task with a corrupt record** — the same breadcrumb tells the model
  to *create a task*, while the real problem is an existing task whose
  `task.json` will not parse. A user debugging "why isn't my task detected"
  gets a confident, wrong answer, and the model may be nudged toward creating a
  duplicate task rather than surfacing the broken file.

The failure is silent in the sense that matters: nothing anywhere reports that
a `task.json` failed to parse.

### Suggested fix

Either direction closes the contradiction, but they are not equally useful:

1. **Correct the behavior** (preferred) — distinguish "no task" from "task
   record unreadable". Even a distinct breadcrumb saying the active task's
   record could not be parsed would turn a misleading answer into an actionable
   one. Note that genuine silence is also wrong here; the docstring's own
   promise of no output would leave a corrupt `task.json` completely
   undiagnosed.
2. **Correct the docstring** — drop the second bullet and state that a
   malformed or status-less `task.json` is reported as `no_task`. Cheap, but it
   documents the confusing behavior rather than fixing it.

### Verification

Read from `main` via the contents API rather than a vendored copy; line numbers
above are from that file (475 lines). This is a static read of the code path —
I did not execute the hook, since reproducing it requires the session-aware
active-task resolver it imports.

Not a duplicate of #356, which was a stdin-piping hang in the same file.

izumi0uu:
## Summary

Fix the workflow-state hook so an existing active task with an unreadable task record is not reported as `no_task`.

When the session pointer resolves to a task directory whose `task.json` is missing, malformed, not an object, or has no usable `status`, the hook now emits `Task: <task> (task_error)`. This preserves the task identity and gives the model a diagnostic state instead of encouraging duplicate task creation.

The behavior is synchronized across the shared template, Claude hook, and Codex hook. The runtime contract now documents `task_error` as a reachable pseudo-status.

Closes #544

## Testing

- [x] Core test suite: 344 passed, 1 skipped
- [x] CLI test suite: 1,674 passed
- [x] TypeScript typecheck
- [x] ESLint
- [x] Python basedpyright: 0 errors (existing unused-import warnings only)
- [x] Python syntax compilation
- [x] GitNexus change detection: low risk
- [x] `git diff --check`

## Scope

- [x] Hooks and CLI regression tests
- [ ] Backend / Lambda
- [ ] Frontend
- [ ] Infrastructure
- [ ] Database migration

## Regression Coverage

- Malformed JSON
- Missing or unusable `status`
- Non-object JSON task records
- Ensures the result does not contain `Status: no_task`

## Summary by CodeRabbit

* **Bug Fixes**
  * Workflow-state reporting now distinguishes task-record errors from sessions without an active task.
  * Displays a diagnostic `task_error` status when task details are missing, unreadable, malformed, non-object, or incomplete.
  * Preserves the task directory name in error breadcrumbs and provides guidance to repair the existing task record without creating a replacement.

* **Tests**
  * Added regression coverage for invalid task records and repair guidance in workflow messages.

### DIFF
--- a/.claude/hooks/inject-workflow-state.py
+++ b/.claude/hooks/inject-workflow-state.py
@@ -27,9 +27,12 @@ custom agent's ``hooks.userPromptSubmit`` and the IDE ``.kiro.hook``
 ``promptSubmit`` event; its output branch emits a plain-text breadcrumb
 (Kiro adds hook stdout directly to the conversation context).
 
-Silent exit 0 cases (no output):
+Silent exit 0 case (no output):
   - No .trellis/ directory found (not a Trellis project)
-  - task.json malformed or missing status
+
+When a session points at a task directory whose task.json is missing, malformed,
+or missing a usable status, the hook emits a task_error breadcrumb instead of
+misreporting the session as having no active task.
 """
 from __future__ import annotations
 
@@ -155,8 +158,16 @@ def _resolve_active_task(root: Path, input_data: dict):
     return resolve_active_task(root, input_data, platform=_detect_platform(input_data))
 
 
-def get_active_task(root: Path, input_data: dict) -> Optional[tuple[str, str, str]]:
-    """Return (task_id, status, source) from the current active task."""
+def get_active_task(
+    root: Path, input_data: dict
+) -> tuple[str, str, str] | None:
+    """Return active task data, a task-record error, or no task pointer.
+
+    ``(task_id, "task_error", source)`` is distinct from ``None``: a session
+    pointer can exist even when its task record is missing or unreadable, and
+    that state needs a diagnostic breadcrumb rather than the normal ``no_task``
+    prompt.
+    """
     active = _resolve_active_task(root, input_data)
     if not active.task_path:
         return None
@@ -169,16 +180,18 @@ def get_active_task(root: Path, input_data: dict) -> Optional[tuple[str, str, st
 
     task_json = task_dir / "task.json"
     if not task_json.is_file():
-        return None
+        return task_dir.name, "task_error", active.source
     try:
         data = json.loads(task_json.read_text(encoding="utf-8"))
     except (json.JSONDecodeError, OSError):
-        return None
+        return task_dir.name, "task_error", active.source
+    if not isinstance(data, dict):
+        return task_dir.name, "task_error", active.source
 
     task_id = data.get("id") or task_dir.name
     status = data.get("status", "")
     if not isinstance(status, str) or not status:
-        return None
+        return task_dir.name, "task_error", active.source
     return task_id, status, active.source
 
 
--- a/.codex/hooks/inject-workflow-state.py
+++ b/.codex/hooks/inject-workflow-state.py
@@ -27,9 +27,12 @@ custom agent's ``hooks.userPromptSubmit`` and the IDE ``.kiro.hook``
 ``promptSubmit`` event; its output branch emits a plain-text breadcrumb
 (Kiro adds hook stdout directly to the conversation context).
 
-Silent exit 0 cases (no output):
+Silent exit 0 case (no output):
   - No .trellis/ directory found (not a Trellis project)
-  - task.json malformed or missing status
+
+When a session points at a task directory whose task.json is missing, malformed,
+or missing a usable status, the hook emits a task_error breadcrumb instead of
+misreporting the session as having no active task.
 """
 from __future__ import annotations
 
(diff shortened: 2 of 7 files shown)


taosu0216:
Reviewed — the fix is right and I want to merge it, with one gap worth closing first.

## The fix

Distinguishing "the record cannot be read" from "there is no task" is the correct call, and the failure mode it prevents is the expensive one: the model was being told `no_task` and would go create a duplicate task on top of a real one it simply could not read. The three hook copies (`.claude/`, `.codex/`, `packages/cli/src/templates/shared-hooks/`) are byte-identical, so the dogfood copies will not drift.

It also does not make the hook noisier — `UserPromptSubmit` already runs every turn, so this relabels an existing per-turn line rather than adding a new trigger.

## The gap

The PR adds the `task_error` status but not a `[workflow-state:task_error]` block in `packages/cli/src/templates/trellis/workflow.md`. I verified the template has no such block today, and `build_breadcrumb` falls back to the generic `"Refer to workflow.md for current step."` line when a tag is missing.

So after merging, a user with a corrupt `task.json` sees:

```
Task: <id> (task_error)
Refer to workflow.md for current step.
```

The header — the part that stops the duplicate-task behaviour — works. But the diagnostic body your description promises never appears; the specific wording in the test (`Repair the active task record before continuing.`) comes from a `workflow.md` the test writes itself, so it does not demonstrate what ships.

The fallback is deliberate design (the docstring is explicit that workflow.md is the only source of truth and a missing tag should degrade visibly rather than be masked), so this is incomplete rather than broken.

Could you add the `[workflow-state:task_error]` block to the bundled template, telling the user their active-task record is unreadable and how to repair it? A few lines. Note the marketplace mirror is byte-compared against the bundled template by `test/templates/trellis.test.ts`, so `marketplace/workflows/native/workflow.md` needs the same edit in a companion PR against `mindfold-ai/marketplace`.

Ping when that is in and I will merge.

taosu0216:
Checking back on this one — I think my earlier request may have read as already satisfied, so to be concrete about what is still missing.

The `[workflow-state:task_error]` blocks in this PR are in `packages/cli/test/regression.test.ts`, where the tests write their own `workflow.md` fixture. The shipped template, `packages/cli/src/templates/trellis/workflow.md`, still has no `task_error` block — I checked against the current branch head (`fd25b407`).

So the tests demonstrate the hook renders the block correctly *if one exists*, but no such block ships. A real user with a corrupt `task.json` gets the header (`Task: <id> (task_error)`) and then the generic `Refer to workflow.md for current step.` fallback.

What is needed is a `[workflow-state:task_error]` block in the bundled template itself, saying the active-task record could not be read and how to repair it. Then the marketplace mirror needs the same edit in a companion PR against `mindfold-ai/marketplace`, because `test/templates/trellis.test.ts` compares `marketplace/workflows/native/workflow.md` byte-for-byte against the bundled template and will fail otherwise.

The rest of the PR is good and I want it in — the hook change is correct, the three copies are byte-identical, and stopping the duplicate-task behaviour is the valuable part. This is the last piece.

izumi0uu:
Thanks, I forgot to follow up. I’ve added the missing shipped `[workflow-state:task_error]` block.

It now tells the agent to repair the existing active task record and explicitly prevents creating or activating another task. I also updated the workflow status map/table and added a template-level regression assertion, so this behavior is now verified against the bundled workflow rather than only a test fixture.

The native marketplace mirror is byte-identical and is available in the companion PR:
https://github.com/mindfold-ai/marketplace/pull/15

Local verification:
- Core: 344 passed, 1 skipped
- CLI: 1693 passed
- Bundled/marketplace workflow parity confirmed

taosu0216 (approved):
Looks good: unreadable `task.json` is now a `task_error` breadcrumb instead of a false `no_task`, with regression coverage and native workflow guidance.

Note for a follow-up (not blocking): marketplace `workflows/tdd/workflow.md` did not get the matching `task_error` block. Native is updated.

PR #547, opened 2026-08-13
feat(cli): add DeepSeek Harness (dsh) as an init platform

taosu0216:
## Summary

Adds **DeepSeek Harness (dsh)** as a Trellis platform: `trellis init --dsh` installs the Trellis workflow into a dsh project.

dsh is a skills-first, class-2 pull-based host:

- it discovers `<projectRoot>/.agents/skills` (rank 200) and `<projectRoot>/.dsh/skills` (rank 100) natively, and the agent loads skills by name through its skill-loader tool;
- the default `web` / `headless` profiles ship no session-start hook, so `hasHooks: false` and `trellis-start` stays a user-invocable skill;
- there is no project-level sub-agent definition surface, so no `trellis-implement` / `trellis-check` / `trellis-research` agent prompts are written — those phases run inline through the workflow skills.

## File map installed by `--dsh`

| Path | Content |
| --- | --- |
| `.agents/skills/` | workflow skills (`trellis-before-dev`, `trellis-brainstorm`, `trellis-check`, `trellis-break-loop`, `trellis-update-spec`) + bundled skills — written via the neutral resolver, **byte-identical** to the Codex / Gemini CLI / Pi / Kimi writes into the same shared root |
| `.dsh/skills/` | dsh-private entry skills (`trellis-start` / `trellis-continue` / `trellis-finish-work`), platform-resolved (`--platform dsh`, bare `trellis-<name>` skill refs) |
| `.dsh/DSH.md` | operator guide |

## Registration

- `AITool` / `TemplateDir` / `CliFlag` unions + `AI_TOOLS.dsh` entry (`configDir: ".dsh"`, `supportsAgentSkills: true`, `hasHooks: false`, `agentCapable: true`)
- new `cmdRefPrefix` member `"trellis-"` — dsh references other skills by bare name
- `configurators/dsh.ts` + `templates/dsh/` + `PLATFORM_FUNCTIONS` entry
- `--dsh` CLI flag + `InitOptions.dsh`
- docs: platform-map.md, README/README_CN platform count 21 → 22

## Verification

- `pnpm typecheck` ✅, `pnpm lint` ✅
- full CLI test suite: **1677/1677 pass** (74 files), including new `test/templates/dsh.test.ts` (5 tests), a `configurePlatform('dsh')` integration test, `.dsh` managed-path cases, and the generic per-platform invariants that now cover dsh
- end-to-end in a scratch repo: `trellis init --dsh -y -u test` → `trellis platforms` reports `DeepSeek Harness (dsh)`, `trellis update --dry-run` is idempotent, `trellis init --help` shows `--dsh`

## Follow-ups

- docs-site (separate submodule repo) — the "Supported Platforms" page that lists init flags needs `--dsh` added; the registry-vs-docs test runs in CI and will flag it there.
- The dsh-side skill frontmatter (`name` + `description`) matches dsh's skill-filesystem parser requirements; a final dsh-boot smoke in an initialized project is planned separately.

## Summary by CodeRabbit

* **New Features**
  * Added DeepSeek Harness support to `trellis init` through the `--dsh` option.
  * Setup now generates shared workflow skills, DeepSeek Harness-specific skills, and an operator guide.
  * Added DeepSeek Harness compatibility across Trellis workflows, including task tracking and platform-specific skill loading.
  * Updated supported platform documentation and capability listings.

* **Documentation**
  * Updated platform coverage from 21 to 22 AI coding platforms in English and Chinese documentation.


SajoLuo:
Thanks for opening this PR. I did a second, implementation-level review against DSH `0.1.0-rc.6`, then built and exercised a fuller native integration. I would like to coordinate before opening another PR so we do not create a competing implementation or mix the `main` and `v0.7` release lines.

## What we learned from a real DSH run

The initial skills-only adapter is a useful baseline, but DSH can preserve more of the Trellis experience than we first assumed:

- DSH has continuable native subagents and exposes each agent's native `DSH_SESSION_ID` to its managed shell.
- Project-local Trellis role instructions can be delivered as non-user-invocable, collision-free skills:
  `trellis-agent-research`, `trellis-agent-implement`, and `trellis-agent-check`.
- The parent can dispatch those roles in DSH's default background mode and synchronize on the native `subagent/end` settlement event instead of blocking in the foreground or polling.
- DSH sessions may inherit an outer `CODEX_THREAD_ID`; platform detection therefore has to prefer `DSH_SHELL=1` + `DSH_SESSION_ID` so Trellis writes a `dsh_<session-id>.json` task pointer rather than a foreign Codex pointer.

## Current design

The current design has two layers.

### Trellis adapter

- Shared workflow and bundled skills remain under `.agents/skills/`, rendered through the neutral resolver.
- DSH-private entry skills and role skills live under `.dsh/skills/`.
- Role skills use `user-invocable: false` and distinct `trellis-agent-*` names so they do not shadow main-session Trellis skills.
- Implement/check roles keep the pull-based context prelude; every child is dispatched with an `Active task:` first line and loads exactly one role skill.
- The workflow defaults to continuable background children, allows the main agent to do independent work, and explicitly forbids shell sleep, polling loops, `job_output`, and repeated agent-list polling.
- Session identity handling prefers the native managed DSH shell over inherited outer-host variables.
- Project dogfood scripts and installed templates are kept in sync, with regression coverage for platform precedence, template hashes, init behavior, and role frontmatter.

### DSH host plugin

The host-side functionality is intentionally kept in a separate plugin rather than embedded in Trellis:

- per-turn `[workflow-state:*]` breadcrumb injection;
- native DSH session identity forwarding for plugin commands;
- read-only `/trellis-status` and `/trellis-finish` helpers;
- an event-driven `trellis_wait` tool that validates a direct continuable child, listens for its scoped `subagent/end`, handles cancellation/disposal, and returns only after DSH has queued the native settlement notice.

The plugin is now public at:
https://github.com/SajoLuo/dsh-trellis

One contribution question is whether upstream is comfortable documenting this as an optional companion. `trellis init --dsh` does not install a profile plugin itself, so before an upstream PR I would make the workflow capability-aware: use `trellis_wait` when available and fall back to an initial foreground dispatch when it is not, without introducing polling.

## Published branches

The complete Trellis work is currently based on `feat/v0.7-beta` / `0.7.0-beta.3`, not `main`:

- Trellis adapter and tests:
  https://github.com/SajoLuo/Trellis/tree/agent/dsh-native-integration
- docs submodule updates:
  https://github.com/SajoLuo/docs/tree/agent/dsh-platform-docs
- marketplace native-workflow mirror:
  https://github.com/SajoLuo/marketplace/tree/agent/dsh-native-workflow
- DSH plugin:
  https://github.com/SajoLuo/dsh-trellis

The docs and marketplace changes are committed in their own forks. The Trellis root branch intentionally does not record fork-only submodule SHAs; those pointers should move only after the child commits are available from the upstream submodule repositories.

## Verification

Trellis, in a read-only-source Linux Docker validation as the non-root `node` user:

- build: pass
- core: 344 passed, 1 skipped
- CLI: 1802 passed, 2 skipped
- DSH template tests: 8 passed
- core + CLI lint: pass
- TypeScript typecheck: pass
- basedpyright: 0 errors (64 existing unused-import warnings)

Plugin:

- Windows: 29/29 tests passed
- Linux `node:24-bookworm`, non-root: 29/29 passed
- `npm pack --dry-run`: pass, 11 published files including the MIT license

Real end-to-end DSH run:

- initialized a clean fixture with the built `trellis init --dsh`;
- completed research -> implement -> check -> spec update -> commit -> task archive -> journal;
- three native child dispatches and three event-driven `trellis_wait` calls;
- zero `Start-Sleep`, shell `sleep`, `job_output`, or `list_agents` polling;
- correct `dsh_<session-id>` runtime pointer even with an inherited outer Codex environment;
- fixture tests: 9/9 under normal Windows Node isolation and 9/9 in Linux Docker;
- final fixture worktree clean, with no active task left behind.

The first diagnostic run also completed successfully, but it exposed the two issues above: inherited Codex identity and foreground-like sleep/poll waiting. The second run verifies both fixes.

## How would you prefer to proceed?

Because this PR targets `main` while the fuller implementation uses the v0.7 task/session machinery, which direction would be easiest for the maintainers?

1. Keep #547 as the minimal `main` adapter, then submit the native session/subagent/plugin-assisted behavior as a follow-up against `feat/v0.7-beta`;
2. Port the full implementation back onto #547/`main`;
3. Supersede or retarget #547 with the fuller v0.7 implementation.

Also, should I open the docs and marketplace PRs now, or wait until the desired Trellis base and plugin fallback contract are agreed?

I am happy to reorganize the commits or provide focused cherry-picks once the preferred direction is clear.

taosu0216:
> Thanks for opening this PR. I did a second, implementation-level review against DSH `0.1.0-rc.6`, then built and exercised a fuller native integration. I would like to coordinate before opening another PR so we do not create a competing implementation or mix the `main` and `v0.7` release lines.
> 
> ## What we learned from a real DSH run
> The initial skills-only adapter is a useful baseline, but DSH can preserve more of the Trellis experience than we first assumed:
> 
> * DSH has continuable native subagents and exposes each agent's native `DSH_SESSION_ID` to its managed shell.
> * Project-local Trellis role instructions can be delivered as non-user-invocable, collision-free skills:
>   `trellis-agent-research`, `trellis-agent-implement`, and `trellis-agent-check`.
> * The parent can dispatch those roles in DSH's default background mode and synchronize on the native `subagent/end` settlement event instead of blocking in the foreground or polling.
> * DSH sessions may inherit an outer `CODEX_THREAD_ID`; platform detection therefore has to prefer `DSH_SHELL=1` + `DSH_SESSION_ID` so Trellis writes a `dsh_<session-id>.json` task pointer rather than a foreign Codex pointer.
> 
> ## Current design
> The current design has two layers.
> 
> ### Trellis adapter
> * Shared workflow and bundled skills remain under `.agents/skills/`, rendered through the neutral resolver.
> * DSH-private entry skills and role skills live under `.dsh/skills/`.
> * Role skills use `user-invocable: false` and distinct `trellis-agent-*` names so they do not shadow main-session Trellis skills.
> * Implement/check roles keep the pull-based context prelude; every child is dispatched with an `Active task:` first line and loads exactly one role skill.
> * The workflow defaults to continuable background children, allows the main agent to do independent work, and explicitly forbids shell sleep, polling loops, `job_output`, and repeated agent-list polling.
> * Session identity handling prefers the native managed DSH shell over inherited outer-host variables.
> * Project dogfood scripts and installed templates are kept in sync, with regression coverage for platform precedence, template hashes, init behavior, and role frontmatter.
> 
> ### DSH host plugin
> The host-side functionality is intentionally kept in a separate plugin rather than embedded in Trellis:
> 
> * per-turn `[workflow-state:*]` breadcrumb injection;
> * native DSH session identity forwarding for plugin commands;
> * read-only `/trellis-status` and `/trellis-finish` helpers;
> * an event-driven `trellis_wait` tool that validates a direct continuable child, listens for its scoped `subagent/end`, handles cancellation/disposal, and returns only after DSH has queued the native settlement notice.
> 
> The plugin is now public at: https://github.com/SajoLuo/dsh-trellis
> 
> One contribution question is whether upstream is comfortable documenting this as an optional companion. `trellis init --dsh` does not install a profile plugin itself, so before an upstream PR I would make the workflow capability-aware: use `trellis_wait` when available and fall back to an initial foreground dispatch when it is not, without introducing polling.
> 
> ## Published branches
> The complete Trellis work is currently based on `feat/v0.7-beta` / `0.7.0-beta.3`, not `main`:
> 
> * Trellis adapter and tests:
>   https://github.com/SajoLuo/Trellis/tree/agent/dsh-native-integration
> * docs submodule updates:
>   https://github.com/SajoLuo/docs/tree/agent/dsh-platform-docs
> * marketplace native-workflow mirror:
>   https://github.com/SajoLuo/marketplace/tree/agent/dsh-native-workflow
> * DSH plugin:
>   https://github.com/SajoLuo/dsh-trellis
> 
> The docs and marketplace changes are committed in their own forks. The Trellis root branch intentionally does not record fork-only submodule SHAs; those pointers should move only after the child commits are available from the upstream submodule repositories.
> 
> ## Verification
> Trellis, in a read-only-source Linux Docker validation as the non-root `node` user:
> 
> * build: pass
> * core: 344 passed, 1 skipped
> * CLI: 1802 passed, 2 skipped
> * DSH template tests: 8 passed
> * core + CLI lint: pass
> * TypeScript typecheck: pass
> * basedpyright: 0 errors (64 existing unused-import warnings)
> 
> Plugin:
> 
> * Windows: 29/29 tests passed
> * Linux `node:24-bookworm`, non-root: 29/29 passed
> * `npm pack --dry-run`: pass, 11 published files including the MIT license
> 
> Real end-to-end DSH run:
> 
> * initialized a clean fixture with the built `trellis init --dsh`;
> * completed research -> implement -> check -> spec update -> commit -> task archive -> journal;
> * three native child dispatches and three event-driven `trellis_wait` calls;
> * zero `Start-Sleep`, shell `sleep`, `job_output`, or `list_agents` polling;
> * correct `dsh_<session-id>` runtime pointer even with an inherited outer Codex environment;
> * fixture tests: 9/9 under normal Windows Node isolation and 9/9 in Linux Docker;
> * final fixture worktree clean, with no active task left behind.
> 
> The first diagnostic run also completed successfully, but it exposed the two issues above: inherited Codex identity and foreground-like sleep/poll waiting. The second run verifies both fixes.
> 
> ## How would you prefer to proceed?
> Because this PR targets `main` while the fuller implementation uses the v0.7 task/session machinery, which direction would be easiest for the maintainers?
> 
> 1. Keep [feat(cli): add DeepSeek Harness (dsh) as an init platform #547](https://github.com/mindfold-ai/Trellis/pull/547) as the minimal `main` adapter, then submit the native session/subagent/plugin-assisted behavior as a follow-up against `feat/v0.7-beta`;
> 2. Port the full implementation back onto [feat(cli): add DeepSeek Harness (dsh) as an init platform #547](https://github.com/mindfold-ai/Trellis/pull/547)/`main`;
> 3. Supersede or retarget [feat(cli): add DeepSeek Harness (dsh) as an init platform #547](https://github.com/mindfold-ai/Trellis/pull/547) with the fuller v0.7 implementation.
> 
> Also, should I open the docs and marketplace PRs now, or wait until the desired Trellis base and plugin fallback contract are agreed?
> 
> I am happy to reorganize the commits or provide focused cherry-picks once the preferred direction is clear.

I think option `1` is better, thank you for the contribution! As for the related bug you found( `CODEX_THREAD_ID` ), I will make new commits in this PR to address them.
You can submit a PR to beta branch, and after you submit the PR, it will be remerged into the main branch after I merge it.

SajoLuo:
Thanks — option 1 works for me.

I’ll keep #547 as the minimal `main` integration and prepare a focused follow-up PR against `feat/v0.7-beta`. I’ll leave the `main`-side `CODEX_THREAD_ID` fix to this PR; the beta implementation will retain the equivalent DSH session-precedence behavior where needed.

Before opening the beta PR, I’ll finish the remaining integration cleanup: make the workflow capability-aware so `trellis_wait` is used only when the companion plugin is available, with a foreground dispatch fallback otherwise; remove fork-only submodule pointers and keep the marketplace mirror reproducible; align the DSH research template with Trellis’s `python3`/Windows rendering convention; and complete the platform spec/docs/test synchronization. I’ll then rerun the clean Linux/Windows verification and open the Trellis beta PR first, with the linked docs/marketplace companion work following as appropriate.

Thanks for coordinating this.

taosu0216:
Merged and shipped in `0.6.15` — thanks again for the review depth here.

## What landed on `main`

The `CODEX_THREAD_ID` fix is in the same release. `dsh` now sits **first** in `_ENV_SESSION_KEYS` in `.trellis/scripts/common/active_task.py`. The root cause was slightly more mechanical than "detection has to prefer DSH": since dsh ships no hook, `task.py` resolves through the shell path with no platform hint, and `_lookup_env_context_key` then walks the whole table in order — so whichever vendor name matched first claimed the session. `DSH_SESSION_ID` is a name no other vendor sets, so first place is safe and cannot mis-claim a non-dsh session.

The regression case pins the ordering rather than the lookup: it sets `DSH_SESSION_ID` and `CODEX_THREAD_ID` together with no platform hint and asserts `dsh_own`. I verified it actually fails (`codex_outer`) when the entry is moved down the table, so it will catch a future reordering rather than just passing vacuously. You are credited as co-author on the squashed commit.

## One thing that will affect your beta PR

Your report surfaced a second gap indirectly. `workflow.md` splits its routing sections into a sub-agent-dispatch block and an inline-execution block, and `dsh` was in **neither** — an agent reading `workflow.md` on dsh found no routing block it belonged to. On `main` I put it in the inline group, since the minimal adapter has no sub-agent surface:

```
[codex-inline, Kilo, Antigravity, Devin, DeepSeek Harness]
```

**Your native implementation inverts this.** With continuable native subagents and `trellis_wait`, dsh belongs in the sub-agent-dispatch block on the beta line, not the inline one — so expect to move it across rather than inherit the `main` placement when you rebase. Worth flagging now because it is a semantic conflict that a clean textual merge will not surface.

Two related notes: the marketplace mirror is enforced byte-identical to the bundled template by `test/templates/trellis.test.ts`, so any `workflow.md` change has to be mirrored in the same PR; and the block markers are consumed by the reading agent, not by a build step, so a platform missing from every block silently degrades rather than failing a test.

## docs and marketplace

Please go ahead and open both now — neither repo has a beta line, both are just `main`.

I have already covered the stable side for the minimal adapter in `0.6.15`: platform count 21 → 22 across EN and ZH, `--dsh` added to every `Supported flags` list, a `DeepSeek Harness` entry under "Other supported platforms", and the sub-agent count left at 18. So scope your docs PR to the `beta/` pages describing the native integration and we will not collide. `registry-invariants.test.ts` scans the whole docs site and will tell you immediately if a page is left behind.

Happy to document `dsh-trellis` as an optional companion, on the contract you already proposed: `trellis_wait` when the plugin is present, foreground dispatch fallback when it is not, no polling either way. Keeping it a separate plugin is the right call — `trellis init --dsh` should not be installing a profile plugin.

Your remaining cleanup list matches what I would have asked for, including the `python3`/Windows rendering convention. No need to reorganize commits on our account; open the beta PR whenever the verification run is clean.

taosu0216:
Follow-up: I ran `0.6.15` against a real dsh `0.1.0-rc.6` on macOS to verify the shipped adapter end to end, and it surfaced one more inheritance path worth flagging before you finalize the beta PR — this one is upstream of the `CODEX_THREAD_ID` fix and will bite the plugin harder than it bites `main`.

## Verified working

`trellis init --dsh` into a clean repo, then from inside a real dsh session:

- both skill roots are discovered natively with no registration step — `trellis-start` / `trellis-continue` / `trellis-finish-work` from `.dsh/skills/`, and the bundled skills from `.agents/skills/`;
- `task.py create` → `start` → `current` completes, and the runtime pointer lands at `.trellis/.runtime/sessions/dsh_session-<id>.json` with the correct prefix.

Your report about the managed shell also checks out: `DSH_SESSION_ID`, `DSH_SHELL` and `DSH_SESSION_JSONL` all reach shell children.

I also reproduced the inherited-identity case directly, with Claude Code as the outer host instead of Codex — the dsh shell carried both `DSH_SESSION_ID` and an inherited `CLAUDE_CODE_SESSION_ID`, and resolution correctly returned `dsh_session-<id>`. So the table-ordering fix holds in a live environment, not just in the unit test.

## The new one: `TRELLIS_CONTEXT_ID` short-circuits everything

My first probe returned `claude_<outer-session-id>` instead of a dsh key. The fix had not failed — `resolve_context_key` checks `TRELLIS_CONTEXT_ID` **before** it ever reaches `_ENV_SESSION_KEYS`:

```python
if allow_environment_context:
    override = _string_value(os.environ.get("TRELLIS_CONTEXT_ID"))
    if override:
        return _sanitize_key(override) or _hash_value(override)
```

My outer Claude Code session had `TRELLIS_CONTEXT_ID` set by its own Trellis hook, dsh inherited it like any other env var, and the override won. Unsetting it alone flipped the result to `dsh_session-<id>`.

This is by design as an explicit override — the problem is that it is inherited indiscriminately by child processes. Concretely: **launch dsh from any already-active Trellis session (Claude Code, Codex, anything hook-backed) and the dsh session silently adopts the outer session's active task.** No env var named after a vendor is involved, so the precedence work in `_ENV_SESSION_KEYS` cannot help here.

Two reasons this matters more for your branch than for `main`:

1. your plugin forwards native dsh session identity for plugin commands — if the forwarding path resolves through `resolve_context_key` rather than reading `DSH_SESSION_ID` directly, an inherited override outranks the identity you are deliberately forwarding;
2. it is invisible rather than wrong-looking. The task pointer resolves, `task.py current` answers, work gets attributed to a session that is not the one running it. My probe only caught it because the returned id happened to match a directory name I recognized.

I have not fixed this on `main` — it is a pre-existing behavior that predates dsh and changing override semantics deserves its own discussion, not a drive-by patch inside a platform PR. Flagging it so your capability-aware work can decide whether the plugin should scrub or override `TRELLIS_CONTEXT_ID` when it establishes a dsh session.

## Minor, for your test matrix

Each `dsh --profile headless` invocation is a fresh session (two consecutive probes produced different session ids), so a task pointer does not persist across headless calls. Fine for the interactive profiles, but worth knowing if any fixture test drives dsh headlessly and expects an active task to survive between invocations.

SajoLuo:
Thanks for the live macOS follow-up ? the `TRELLIS_CONTEXT_ID` short-circuit was real, and I incorporated it before finalizing the beta follow-up.

The fix is deliberately scoped so it does not change the generic override contract for other platforms:

- `dsh-trellis` now registers a managed `DSH_TRELLIS_CONTEXT_ID = dsh_<session-id>` through DSH's `shellEnv` registry for each shell execution. DSH scrubs ambient `DSH_*` values before rebuilding that namespace, so this identity comes from the current DSH session rather than the outer host.
- The beta resolver checks that plugin-only managed value before the inherited generic `TRELLIS_CONTEXT_ID`.
- Plugin utility-command subprocesses explicitly replace `TRELLIS_CONTEXT_ID` with the same current DSH key.
- Without the optional plugin, the operator/beta docs now say to unset `TRELLIS_CONTEXT_ID` before launching DSH from an active Trellis host; the global explicit-override semantics remain unchanged.
- The regression test sets `TRELLIS_CONTEXT_ID=claude_outer-session`, managed `DSH_TRELLIS_CONTEXT_ID=dsh_inner-session`, and `DSH_SESSION_ID=inner-session`, then asserts only `dsh_inner-session.json` is written.

The coordinated drafts are now open and linked:

- Trellis beta: https://github.com/mindfold-ai/Trellis/pull/548
- Marketplace: https://github.com/mindfold-ai/marketplace/pull/13
- Beta docs: https://github.com/mindfold-ai/docs/pull/31
- Optional plugin: https://github.com/SajoLuo/dsh-trellis (fix at `a73e9ef`)

Fresh non-root `node:24-bookworm` verification from the public branches is clean: build/lint/typecheck pass, basedpyright 0 errors, core 344 passed / 1 skipped, CLI 1805 passed / 2 skipped; the plugin passes 31/31 tests and `npm pack --dry-run`. The root clone also fetched the exact docs and marketplace PR commits from the upstream submodule repositories.

I also documented the headless behavior you observed: each invocation is treated as a fresh session, so tests must not expect an active-task pointer to persist across separate headless runs.
Update after posting: I also reran this as a focused probe in a real DSH `0.1.0-rc.6` headless session. The parent process deliberately had `TRELLIS_CONTEXT_ID=claude_outer-session`; the managed shell exposed a current `dsh_session-*` value, and `resolve_context_key()` returned that exact managed key rather than the outer value. The probe reported PASS.

One operational detail surfaced while refreshing the local `file:` plugin install: pnpm keeps a profile-local snapshot, so a source update that adds a new file requires remove + add for that profile. The packaged artifact already contained the file; I refreshed the profile, reran the live probe successfully, and documented the refresh step in the plugin README (`3c2f53e`).

PR #551, opened 2026-08-14
docs(dsh): clarify bundled meta foreground fallback

SajoLuo:
## Summary

- make the DeepSeek Harness row in the bundled `trellis-meta` platform reference explicit about the no-plugin fallback
- require every DSH child to start with `run_in_background: false` when `trellis_wait` is unavailable
- add a DSH template regression test so the bundled reference cannot drift back to the ambiguous wording

## Why

This is a narrow follow-up to #550 and addresses the unresolved review finding in https://github.com/mindfold-ai/Trellis/pull/550#discussion_r3782080030. The affected row originated in the DSH support merged through #548; no other platform guidance is changed.

## Impact

Generated `trellis-meta` guidance is now consistent with the DSH operator guide and workflow template. Runtime behavior and non-DSH platform documentation are unchanged.

## Validation

- `pnpm --filter @mindfoldhq/trellis test -- test/templates/dsh.test.ts` (10 passed)
- `pnpm --filter @mindfoldhq/trellis exec eslint test/templates/dsh.test.ts`
- `git diff --check`
- GitNexus change analysis: low risk, no affected execution flows

### DIFF
--- a/packages/cli/src/templates/common/bundled-skills/trellis-meta/references/platform-files/agents.md
+++ b/packages/cli/src/templates/common/bundled-skills/trellis-meta/references/platform-files/agents.md
@@ -35,7 +35,7 @@ Agent files should not become generic chat prompts. They should define input sou
 | Reasonix | `.reasonix/skills/trellis-*/SKILL.md` (subagent frontmatter) |
 | ZCode | `.zcode/agents/trellis-*.md` |
 | Kimi Code | `.kimi-code/skills/trellis-*/SKILL.md` (agent prompts as skills, dispatched to the built-in `coder`; research needs its file-editing tools to persist findings) |
-| DeepSeek Harness | `.dsh/skills/trellis-agent-*/SKILL.md` (role prompts as child-only skills; optional `trellis_wait` enables continuable background settlement, otherwise initial foreground dispatch) |
+| DeepSeek Harness | `.dsh/skills/trellis-agent-*/SKILL.md` (role prompts as child-only skills; optional `trellis_wait` enables continuable background settlement; without it, dispatch every child with `run_in_background: false` from the outset) |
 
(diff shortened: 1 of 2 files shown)


taosu0216:
Verified and ready to merge from our side — but first, this one is on me.

I merged #550 having read only the conversation comments, not the inline review threads, so the codex-connector P2 on `DSH.md:42` never reached me. My own check was worse than useless: I grepped for `initially with \`run_in_background\`` and `at initial dispatch` and reported "no ambiguous wording left anywhere", when the bundled meta reference said `otherwise initial foreground dispatch` — a phrasing my pattern could not match. A narrow grep that returns nothing reads exactly like a clean repo. Thanks for picking it up in a follow-up rather than letting it ride.

## Verified

Your test is the right shape — it pins both directions, asserting the explicit wording is present **and** that the ambiguous phrasing is absent, so a revert cannot pass quietly. I confirmed it actually fails when the old wording is restored:

```
× keeps the bundled meta reference explicit about the DSH fallback
  Tests  1 failed | 9 passed (10)
```

Full run on your head:

```
core   344 passed, 1 skipped
cli   1826 passed  (77 files)
lint / tsc  clean
```

## On the remaining phrasings

I swept every `run_in_background` mention across the templates, dogfood and the mirror. Three read differently from the main paragraph and I want to state explicitly that I checked them rather than leaving you to wonder:

- `workflow.md:364` (research) — "dispatch **each** required child", correct, that phase can fan out
- `workflow.md:499` (implement) and `:540` (check) — "dispatch **that** child", also correct: those phases dispatch exactly one child, so the singular is accurate rather than ambiguous

So after this PR the wording is consistent everywhere it needs to be, and deliberately singular in the two places where only one child exists. Nothing further to clean up.

Mark it ready whenever you like and I will merge. No companion PR needed this time — the change is confined to the bundled template, and the marketplace mirror is unaffected.

PR #566, opened 2026-08-21
feat(cli): let an installed pack declare preferred workflow entry points

sdelmas:
## Problem

A wrapper pack that fronts the Trellis workflows with its own commands (e.g. a company command pack wrapping `/trellis:finish-work` behind a finish-work command of its own) has no way to stop Trellis from routing agents back to the wrapped surfaces. The routing literals live in files no consumer can edit — a refresh reverts any local change:

- `workflow.md` — `Flow: … -> /trellis:finish-work` and `trellis-update-spec` routing
- `AGENTS.md` — "prefer `/trellis:finish-work`, `/trellis:continue`"
- `shared-hooks/session-start.py` — `Next-Action: Run /trellis:finish-work`
- `shared-hooks/inject-workflow-state.py` — bootstrap notice pointing at the `trellis-start` skill
- `opencode/lib/session-utils.js` — the same Next-Action line
- `scripts/common/task_store.py` — "Use /trellis:continue …" after task creation

## Change

Add a **declared-entry-point seam**: an installed pack may write `.trellis/entry-points.json` declaring a preferred entry surface per workflow, and the emitters above consult it, falling back to today's literals when nothing is declared.

```json
{
  "schemaVersion": 1,
  "entryPoints": {
    "start": "/sd:sd-start",
    "continue": "/sd:sd-continue",
    "finish-work": "/sd:sd-finish-work",
    "update-spec": "sd-update-spec"
  }
}
```

Trellis never writes the file.

**Write time** — `init` / `update` / `workflow` load the declaration once, and `applyDeclaredEntryPoints()` rides the exact write chokepoints `replacePythonCommandLiterals()` already uses, so every written template carries the declared routing and template hash tracking hashes exactly what it writes (updates keep diffing cleanly, including when the declaration changes between updates).

**Runtime** — the two shared hooks, the OpenCode session lib, and `task_store.py` consult the declaration on each emit.

**Validation is strict and all-or-nothing**: bounded file size, `schemaVersion: 1`, fixed key set, bounded value grammar, and no declared value may embed a replaced literal (keeps substitution idempotent). Any invalid declaration disables the whole seam. Bare skill identities such as the `trellis-start` directory name are deliberately not rewritten — only the four routing tokens `/trellis:start`, `/trellis:continue`, `/trellis:finish-work`, and `trellis-update-spec`.

## Behavior without a declaration

Unchanged, byte for byte — the transform is a no-op and every emitter keeps its historical wording. The full existing test suite (which inits without declarations everywhere) passes untouched.

## Validation

- `pnpm test`: core 344 passed, cli **1723 passed** (includes 16 new entry-point tests: loader validation matrix, substitution, fallback, idempotency)
- `pnpm lint`: clean
- End-to-end: a real `trellis init` over a repo carrying the declaration produces generated markdown with **zero** remaining `/trellis:` literals, substituted `AGENTS.md`, and `task.py create` printing the declared continue entry; the same init without a declaration keeps all literals.

## Provenance

Filed from a downstream consumer where an audit (finding A-005) measured agents being routed to wrapped entry points by vendored files. The pack half of that finding is platypeeps/sd-ai-command-pack#486; this is the Trellis half.

https://claude.ai/code/session_011y3oVg5eZmhF14F5FGmB36

## Summary by CodeRabbit

* **New Features**
  * Added support for optional entry-point declarations, allowing Trellis commands and generated instructions to use customized routes.
  * Applied configured routes consistently across initialization, updates, workflows, templates, and task guidance.
  * Added support for customized start, continue, finish-work, and update-spec actions.
* **Reliability**
  * Invalid, oversized, or unsupported declarations safely fall back to the standard Trellis routes without interrupting command generation.

### DIFF
--- a/.trellis/scripts/common/entry_points.py
+++ b/.trellis/scripts/common/entry_points.py
@@ -1,0 +1,59 @@
+"""Declared entry-point lookup for wrapper packs.
+
+An installed wrapper pack may declare preferred entry surfaces for the core
+workflows in `.trellis/entry-points.json` so routing text points agents at the
+pack's canonical commands instead of the Trellis literals:
+
+    {
+      "schemaVersion": 1,
+      "entryPoints": {
+        "start": "/sd:sd-start",
+        "continue": "/sd:sd-continue",
+        "finish-work": "/sd:sd-finish-work",
+        "update-spec": "sd-update-spec"
+      }
+    }
+
+Trellis never writes the file. Every key is optional, values are bounded
+surface names rendered verbatim, and any invalid shape, key, or value
+disables the whole declaration so callers fall back to the Trellis literal.
+"""
+
+from __future__ import annotations
+
+import json
+import re
+from pathlib import Path
+
+from .paths import DIR_WORKFLOW, get_repo_root
+
+ENTRY_POINT_KEYS = {"start", "continue", "finish-work", "update-spec"}
+_ENTRY_POINT_VALUE_RE = re.compile(r"/?[A-Za-z0-9][A-Za-z0-9/:._-]{0,63}\Z")
+_ENTRY_POINT_MAX_BYTES = 16 * 1024
+
+
+def declared_entry_point(
+    key: str, fallback: str, repo_root: Path | None = None
+) -> str:
+    """Return the declared entry point for `key`, or `fallback`."""
+    try:
+        root = repo_root if repo_root is not None else get_repo_root()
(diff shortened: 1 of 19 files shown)


sdelmas:
Withdrawing: retargeted to the fork that our fleet actually consumes (sdelmas/Trellis#6, rebased onto the fork's main). Happy to reopen or refile here if upstream wants the seam.

PR #569, opened 2026-08-21
fix(channel): preserve UTF-8 across incremental event reads

### LINKED ISSUE #568: Channel watchers corrupt UTF-8 characters split across incremental reads
## Bug Description

The Core and CLI channel watchers decode each newly read buffer before joining it with the previous partial JSONL line. If a read boundary splits a multibyte UTF-8 character, decoding inserts replacement characters. The completed JSON line can still parse, so the watcher silently emits corrupted message text.

The two writes eventually leave a complete, valid JSONL record on disk; the corruption exists only in incremental watcher output. Splitting `中` after its first byte makes the watched event contain `���` while the persisted `events.jsonl` bytes still represent `中`.

## Steps to Reproduce

1. Create a channel and start its watcher with `fromStart: true`.
2. Append a message JSON line in two writes, ending the first write after the first byte of `中`.
3. Let the watcher consume the first chunk, then append the remaining bytes and newline.
4. Read the emitted message event.

Expected: the watcher emits `中`.

Actual: it emits `���`.

Related but distinct from #527: that issue covered a permanently torn final record and append-side sequence allocation. Here the second write completes a valid record and only the incremental watcher output is corrupted.

## Trellis Version

Current `main` at `64e663694201005bc87766ef22de89b8da3d4d79` (`0.6.15`).

## Node.js Version

`v22.23.1`

## Operating System

macOS

cyq1017:
### Problem

The Core and CLI channel watchers decoded each appended buffer independently. A read boundary inside a multibyte UTF-8 character therefore produced replacement characters while leaving the JSON event parseable.

### Change

Keep a streaming `StringDecoder` in each watcher's read state, reset it when the event file disappears or truncates, and cover both live implementations with split-byte regression tests.

### Verification

- Core focused watcher regression: 1 passed
- CLI focused watcher regression: 1 passed
- Core test suite: 347 passed, 1 skipped
- CLI test suite: 1,709 passed
- `pnpm lint`
- `pnpm typecheck`
- focused four-file Prettier check
- `git diff --check`

Closes #568.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed channel event watching so multi-byte UTF-8 characters remain intact when data arrives in partial chunks.
  - Improved handling of event files that are truncated, rotated, replaced, or recreated.

- **Tests**
  - Added coverage confirming complete message text is preserved across incremental reads in both CLI and core channel watchers.

### DIFF
--- a/packages/cli/src/commands/channel/store/watch.ts
+++ b/packages/cli/src/commands/channel/store/watch.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs";
+import { StringDecoder } from "node:string_decoder";
 
 import { eventsPath, channelDir } from "./paths.js";
 import type { ChannelEvent } from "./events.js";
@@ -9,6 +10,7 @@ export type WatchFilter = ChannelEventFilter;
 interface ReadProgress {
   byteOffset: number;
   carry: string;
+  decoder: StringDecoder;
 }
 
 async function readNewEvents(
@@ -20,6 +22,7 @@ async function readNewEvents(
     // the file reappears we'll re-scan it from byte 0.
     state.byteOffset = 0;
     state.carry = "";
+    state.decoder = new StringDecoder("utf8");
     return [];
   }
   const stat = await fs.promises.stat(filePath);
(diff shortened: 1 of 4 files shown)


taosu0216 (approved):
Reviewed: streaming StringDecoder is the right fix for multibyte chars split across read boundaries; decoder reset on file disappearance/truncation is handled; core and CLI watchers fixed symmetrically with regression tests. CI build passed.

PR #574, opened 2026-08-25
feat(mem): restore the OpenCode session reader without a native dependency

sdelmas:
Slice **A** of the #534 resplit, per your landing order. Branched off current `main` (`64e66369`), not off the old branch.

## What this does

`packages/core/src/mem/adapters/opencode.ts` on `main` is still the 34-line silent no-op left behind by the `better-sqlite3` revert. This restores a real reader on the **zero-dependency read-only SQLite parser that already ships in core** (`internal/sqlite-readonly.ts` — unchanged here, and byte-identical between this branch and `main`, which is what makes the slice self-contained).

No native module, no WASM, no install-time build step, so the install-failure regression that forced the revert cannot return.

- name-matched schema validation — a future OpenCode layout change degrades to a structured warning instead of malformed rows
- WAL-consistent snapshots; the database is opened read-only and is byte-identical after a read
- `parent_id` child-session merging for `--include-children`

## Files (6)

| File | Why |
|---|---|
| `core/src/mem/adapters/opencode.ts` | the reader |
| `core/src/mem/internal/paths.ts` | `opencodeDbPath` |
| `core/src/mem/sessions.ts` | dispatch into the restored adapter |
| `core/test/mem/adapters.test.ts`, `core/test/mem/api.test.ts` | coverage |
| `cli/src/commands/mem.ts` | required, not optional: it removes the "OpenCode reader unavailable" notice, which becomes a false statement the moment the reader works. Warning presentation stays a CLI concern. |

## One deviation from your list, flagged for your call

You listed the `configurators/opencode.ts` docstring fix (`2d06433b`) as keep-able. **I left it out**, because its premise does not hold on `main`: upstream's `templates/opencode/package.json` still declares `@opencode-ai/plugin`, so the existing docstring is *accurate* there. The fork's "correction" describes the bare `{"type": "module"}` template produced by the pre-start-gate work — which you excluded from every slice. Applying it here would make the comment wrong. Say the word and I'll add it.

## Scope hygiene

None of the following appear in the diff: `.trellis/workspace/sven/`, task archives, `-sd.N` version identity (both packages stay at `0.6.15`), the marketplace gitlink move, Opus pinning, the pre-start gate.

## Testing

```
pnpm build   → ok
core         → Test Files 19 passed | Tests 371 passed, 1 skipped
cli          → Test Files 76 passed | Tests 1708 passed
```

Note for reviewers running locally: the marketplace submodule must be at this branch's recorded gitlink (`7310a50c`), or `trellis.test.ts > marketplace native workflow mirror` fails spuriously.

B–E follow in your stated order; #534 is being closed with links to the replacements.

## Summary by CodeRabbit

* **New Features**
  * Added read-only support for browsing OpenCode sessions stored in SQLite databases.
  * Added session filtering by workspace and time range, dialogue extraction, compaction summaries, and keyword search.
  * Added support for parent and sub-agent sessions, including optional child-session merging.
  * Added automatic database discovery and support for configurable database locations.

* **Bug Fixes**
  * Improved handling of malformed data, unsupported schemas, unreadable databases, and unstable database files with user-facing warnings.

### DIFF
--- a/packages/cli/src/commands/mem.ts
+++ b/packages/cli/src/commands/mem.ts
@@ -2,8 +2,8 @@
  * mem.ts — CLI wrapper over `@mindfoldhq/trellis-core/mem`.
  *
  * The reusable retrieval / context-extraction logic lives in core; this file
- * owns only CLI concerns: argument parsing, terminal rendering, the OpenCode
- * "reader unavailable" notice, and process exit behavior.
+ * owns only CLI concerns: argument parsing, terminal rendering, warning
+ * presentation, and process exit behavior.
  *
  * Commands:
  *   list                          list sessions (default if no command)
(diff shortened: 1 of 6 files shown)


taosu0216 (approved):
Reviewed: restores the OpenCode reader on the existing zero-dependency SQLite parser (unchanged in this PR), so the native-module install regression cannot return. Name-matched schema validation, per-condition deduped warnings, WAL visibility, hostile-input hardening (bad JSON, out-of-range timestamps) all covered by tests, including a byte-identical-after-read assertion. Ran core suite locally merged with current main: 372/372 pass.

PR #575, opened 2026-08-25
fix(update): repair receipt entries for files already identical to a template

sdelmas:
Slice **B** of the #534 resplit. Branched off current `main` (`64e66369`). Independent of #574 — different files, no ordering constraint between them.

## The bug

`.template-hashes.json` recorded hashes that disagreed with the files written, and omitted entries for files that exist. Either way the receipt stops working as a drift signal: a clean vendored tree reports as locally modified, so genuine customizations can't be told apart from noise. This surfaced during an 8-repo rollout where the receipt was the tool for deciding which local edits were real — it produced false positives on every repo.

## Root cause

`analyzeChanges` classifies a file whose content already equals its template as `unchanged`. The write-back drew only from `newFiles`, `autoUpdateFiles`, and overwritten `changedFiles` — `unchangedFiles` was never written back. So a wrong or absent entry sitting beside an already-correct file could not be repaired by any number of `trellis update` runs.

Worth stating plainly because it inverts the original reasoning: identical content across versions is not what keeps such an entry fresh — it is exactly what freezes it, because the file never leaves the `unchanged` bucket.

An earlier reading blamed key construction for dropping a platform segment. That was disproved: the collector emits the correct hash, and 0.6.7 hashed bytes read from disk. The keys were always right; the repair path was always missing.

## Files (3)

- `cli/src/commands/update.ts` — write back `unchangedFiles` too
- `cli/src/utils/template-hash.ts`
- `cli/test/commands/update.integration.test.ts` — +124 lines

The source commit also carried `.trellis/tasks/08-19-template-hashes-stale/`; dropped per your scrub list.

## Testing

```
pnpm build   → ok
cli          → Test Files 76 passed | Tests 1713 passed   (+5 receipt tests)
core         → Test Files 19 passed | Tests 346 passed, 1 skipped
```

No `.trellis/` paths, no `-sd.N` identity (stays `0.6.15`), no marketplace gitlink.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update repair behavior for unchanged template files with missing or incorrect hash records.
  * Hash records are now repaired during both regular updates and already-up-to-date checks.
  * Files intentionally excluded from hash tracking remain unaffected.
  * Customized files and mixed-ownership files are no longer incorrectly re-hashed.

### DIFF
--- a/packages/cli/src/commands/update.ts
+++ b/packages/cli/src/commands/update.ts
@@ -29,6 +29,7 @@ import {
   removeHash,
   renameHash,
   computeHash,
+  shouldExcludeFromHash,
 } from "../utils/template-hash.js";
 import { compareVersions } from "../utils/compare-versions.js";
 import { toPosix } from "../utils/posix.js";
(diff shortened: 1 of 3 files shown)


taosu0216 (approved):
Reviewed: repairing receipt entries only for files byte-identical to their template cannot bless local edits (customized files land in changedFiles); mixed-ownership paths correctly stay free to drift; the already-up-to-date exit also repairs, which is the exact case that previously froze poisoned entries. Tests cover poisoned/missing/customized/mixed-ownership. CI build passed.

PR #576, opened 2026-08-25
fix(scripts): harden the task script runtime

sdelmas:
Slice **C** of the #534 resplit: the eight fix commits you named, replayed onto current `main` (`64e66369`). No `rename`, no `add_session` rewrite, no workflow.md.

`5a1d59e0` path containment · `c0d7cb7f` create/archive/link collisions · `1cf22b51` JSON read/write failures · `cf8cb25c` lifecycle hooks + config parsing · `e1a17984` empty title/description rejection · `a95e7483` archive `index.lock` retry · `9c88fec5` hook process-tree kill · `c9489ce8` non-list `children` + advisory `stat()` guards

## Conflict resolutions worth your eyes

**Path containment (`5a1d59e0`) — 7 hunks.** `main` had independently landed a weaker version of this check: resolve, then `relative_to(repo_root)`. I took the fork side, which is a strict superset — it resolves *both* operands (so it still covers the symlink case `main`'s version was written for), rejects the tasks directory itself as a target, and requires containment in `tasks_dir` rather than merely `repo_root`. Worth a close look since it's the one place I overrode existing `main` behaviour rather than adding to it.

**`regression.test.ts` (three separate conflicts).** Each incoming side dragged in the "shipped markdown carries no trailing whitespace" block from `e77af366` — your "Also" bucket, not slice C, not on `main`. Took the HEAD side all three times, so that block stays out and lands with whatever slice you eventually want it in.

**Dropped, not resolved:** task artifacts carried along by the picks (`08-08-*`, `08-09-*`, `08-19-*`), and the workflow.md hunks from `e1a17984`.

## One contradiction in the split, and how I handled it

`task-children-normalization.integration.test.ts` arrives with `c9489ce8` — squarely slice C — but it drives `task.py` with `TRELLIS_DEVELOPER: "tester"`. That env var is implemented by `0740d1d6`, which you put in **slice E**, "only after C". So C's own test cannot pass on C as written.

Rather than pull E's feature forward, `setupRepo` now writes `.trellis/.developer` with `name=tester` — which `get_developer()` on current `main` already reads. The `TRELLIS_DEVELOPER` env in `runTask` is left untouched, so the test keeps passing unchanged once E lands. Happy to do it differently if you'd rather the test move to E.

## Scope hygiene

Vendored `.trellis/scripts` and packaged `templates/trellis/scripts` verified byte-identical (`diff -rq -x __pycache__`). No `.trellis/tasks/`, no `.trellis/workspace/sven/`, no workflow.md, no marketplace gitlink, no `-sd.N` (stays `0.6.15`).

Dogfood skill copies under `.agents/`, `.claude/`, `.cursor/`, `.omp/`, `.opencode/`, `.pi/` are included so their docs match the new create-time validation — drop them if you'd rather refresh via `trellis update`.

## Testing

```
cli    → Test Files 77 passed | Tests 1750 passed
core   → Test Files 19 passed | Tests 346 passed, 1 skipped
```

## Summary by CodeRabbit

- **New Features**
  - Task creation now requires a non-empty description and supports `--force` for intentional overwrites.
  - Git operations automatically retry transient index-lock failures.
  - Added clearer diagnostics for invalid task data, configuration, JSON files, and lifecycle actions.

- **Bug Fixes**
  - Improved task path safety, archive collision handling, parent-child linking, and malformed task recovery.
  - Prevented context validation failures when referenced files disappear.
  - Improved lifecycle hook timeout and cleanup behavior.

- **Documentation**
  - Updated task creation examples and guidance to reflect required descriptions and validation rules.


taosu0216 (approved):
Reviewed the flagged path-containment override: the fork side is a strict superset of main's check (resolves both operands, requires containment under tasks_dir, rejects the tasks dir itself) — took a close look and agree with taking it. Also reviewed the hook process-tree kill, index.lock retry, JSON read/write failure reporting, and create/archive collision handling; all sound. Ran the full CLI suite locally on this branch merged with current main (post #584/#569/#581/#575): 1763/1763 pass. Note: workflow.md's create usage line still lacks --description (intentionally out of slice); the new error message self-corrects callers, but worth updating in a follow-up slice.

PR #596, opened 2026-09-01
fix(mem): read SQLite main pages lazily

### LINKED ISSUE #589: fix(mem): avoid loading the entire OpenCode SQLite database into memory
## Summary

`trellis mem --platform opencode` in Trellis 0.6.16 cannot read a healthy multi-gigabyte OpenCode SQLite database because the read-only snapshot implementation loads the entire main DB into one Node buffer before parsing it.

## Environment

- Trellis CLI: `0.6.16`
- OpenCode: `1.18.13`
- macOS, 16 GiB RAM
- OpenCode DB: about 3.3 GiB
- WAL / SHM: about 32 KiB each

## Reproduction

```bash
trellis mem list --platform opencode --global --limit 1
```

Output:

```text
warning: cannot read OpenCode session database (.../opencode.db): cannot capture SQLite snapshot: .../opencode.db
(no sessions)
```

The DB itself is healthy:

```text
PRAGMA page_size       = 4096
PRAGMA page_count      = 875450
PRAGMA freelist_count  = 0
PRAGMA auto_vacuum     = 0
PRAGMA journal_mode    = wal
PRAGMA quick_check     = ok
```

The DB is mostly live OpenCode event data, not reclaimable freelist space. Aggregate measurements:

```text
event table     ~2687.5 MiB
part table       ~382.1 MiB
message table    ~293.6 MiB
sessions              478 (51 root, 427 child)
```

Within the event payload:

```text
message.updated.1       ~1876.6 MiB
message.part.updated.1   ~628.5 MiB
```

This matches OpenCode's known unbounded event-log growth reports, including anomalyco/opencode#33356 and anomalyco/opencode#41175.

## Likely cause

`packages/core/src/mem/internal/sqlite-readonly.ts` currently captures the main database with:

```ts
const mainBytes = fs.readFileSync(mainPath)
```

For a 3.3 GiB live DB, the full-file allocation/snapshot fails before the SQLite reader can list sessions. The CLI reports the outer snapshot error but not the underlying allocation error.

PR #574 restored the OpenCode reader without native dependencies and correctly added WAL-consistent snapshots, but large long-lived OpenCode databases are realistic because the event store has no retention/compaction.

## Expected behavior

The read-only parser should avoid loading the whole main database into memory. Possible approaches:

1. Read SQLite pages on demand with positional `read` calls.
2. Keep only the B-tree/schema/pages required by the queried tables in memory.
3. Apply WAL page overrides lazily.
4. At minimum, detect oversized DBs before `readFileSync` and emit a specific size/memory diagnostic instead of the generic snapshot failure.

The reader must remain read-only and preserve the WAL-consistency guarantees introduced in #574.

## Additional note

`VACUUM` is not a workaround in this case (`freelist_count=0`); deleting user history solely to make `trellis mem` fit in memory would be an undesirable requirement.

cyq1017:
Fixes #589.

The SQLite reader currently loads the whole main database with `readFileSync`, which fails at Node's 2 GiB buffer limit. This keeps a read-only file descriptor for the main database and loads pages on demand, while preserving the captured WAL snapshot and detecting main-file changes.

Verification:
- `pnpm --filter @mindfoldhq/trellis-core test`
- `pnpm --filter @mindfoldhq/trellis-core lint`
- `pnpm typecheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved read-only SQLite database access for large and sparse files without loading the entire database into memory.
  * Added safeguards to detect database changes during reads and preserve consistent snapshots.
  * Improved handling of partial reads and normalized read failures.
  * Ensured resources are properly released when a read-only database is closed.
  * Preserved captured WAL snapshots after subsequent commits.

### DIFF
--- a/packages/core/src/mem/internal/sqlite-readonly.ts
+++ b/packages/core/src/mem/internal/sqlite-readonly.ts
@@ -230,7 +230,8 @@ interface FileStamp {
 }
 
 interface SqliteSnapshot {
-  mainBytes: Uint8Array;
+  mainFd: number;
+  mainStamp: FileStamp;
   walBytes: Uint8Array | null;
   walIndex: WalIndexHeader | null;
 }
@@ -272,21 +273,28 @@ function equalBytes(a: Uint8Array | null, b: Uint8Array | null): boolean {
   return true;
 }
 
+function stampFromStat(stat: fs.Stats): FileStamp {
+  return {
+    size: stat.size,
+    mtimeMs: stat.mtimeMs,
+    ctimeMs: stat.ctimeMs,
+    dev: stat.dev,
+    ino: stat.ino,
+  };
+}
+
 function fileStamp(path: string): FileStamp | null {
   try {
-    const stat = fs.statSync(path);
-    return {
-      size: stat.size,
-      mtimeMs: stat.mtimeMs,
-      ctimeMs: stat.ctimeMs,
-      dev: stat.dev,
-      ino: stat.ino,
-    };
+    return stampFromStat(fs.statSync(path));
   } catch (error) {
     const code = (error as NodeJS.ErrnoException).code;
     if (code === "ENOENT") return null;
     throw error;
   }
+}
+
+function fileDescriptorStamp(fd: number): FileStamp {
+  return stampFromStat(fs.fstatSync(fd));
 }
 
 function sameStamp(a: FileStamp | null, b: FileStamp | null): boolean {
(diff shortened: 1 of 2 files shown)


taosu0216 (approved):
Reviewed lazy page-read approach. Replaces readFileSync with fd + positional page reads; WAL snapshot preserved; main-file stamp checks; no native sqlite. CI build SUCCESS. Approving.

